PR #14, opened 2017-03-02
[sabayon] Fix use of armv7l arch and setting of ssh service through systemd

geaaru:
Hi,
fix build of armv7 image.

### DIFF
--- a/templates/sabayon.json
+++ b/templates/sabayon.json
@@ -1,16 +1,15 @@
 {
     "create_message": ["You just created a Sabayon container ",
-                       "(release=RELEASE, arch=ARCH, variant=VARIANT)\n\n",
-                       "To enable sshd, run: rc-update add sshd"],
+                       "(release=RELEASE, arch=ARCH)\n\n",
+                       "To enable sshd, run: systemctl enable sshd"],
     "expiry": 30,
     "templates": ["/etc/conf.d/hostname", "/etc/hostname", "/etc/hosts"],
     "config_user": ["common", "userns"],
     "config_system": ["common"],
     "template_arch": {"amd64": "amd64",
-                      "armhf": "arm",
-                      "i386": "x86"},
-    "template_args": ["-a", "ARCH"],
+                      "armhf": "armv7l"},
+    "template_args": ["-a", "ARCH", "-u"],
     "template_env": {"root_password": "root"},
     "post_create": ["passwd -dl root",
-                    "rc-update del sshd"]
+                    "systemctl disable sshd"]
 }


geaaru:
Hi, I don't know if currently are supported template params without param value. On lxc-sabayon template  is possible use "-u" argument for build of unprivileged container.
I'm waiting for a feedback.
Thanks
G.

templates/sabayon.json
```diff
@@ -1,16 +1,15 @@
 {
     "create_message": ["You just created a Sabayon container ",
-                       "(release=RELEASE, arch=ARCH, variant=VARIANT)\n\n",
-                       "To enable sshd, run: rc-update add sshd"],
+                       "(release=RELEASE, arch=ARCH, u=Unprivileged Container)\n\n",
```
stgraber: That part doesn't make sense. The lxc-download template can generate both privileged and unprivileged containers.

templates/sabayon.json
```diff
@@ -1,16 +1,15 @@
...
+                       "To enable sshd, run: systemctl enable sshd"],
     "expiry": 30,
     "templates": ["/etc/conf.d/hostname", "/etc/hostname", "/etc/hosts"],
     "config_user": ["common", "userns"],
     "config_system": ["common"],
     "template_arch": {"amd64": "amd64",
-                      "armhf": "arm",
-                      "i386": "x86"},
+                      "armv7l": "armv7l"},
```
stgraber: that should be

    "armhf": "armv7l"

templates/sabayon.json
geaaru: Yes I know but on lxc-sabayon script I added an option -u because I permit create of unprivileged container directly from lxc script (different of others distro scripts) and with this option I execute some other stuff for systemd to correctly start of sabayon distro on unprivileged container.
See https://github.com/lxc/lxc/blob/master/templates/lxc-sabayon.in#L193

This means that could be correct create two container images for sabayon: one for privileged container and one for unprivileged.

WDYT?

templates/sabayon.json
geaaru: ok, i will fix this.

templates/sabayon.json
stgraber: Ah, then the image we build on Jenkins should always use "-u" so that the resulting image can be used for both privileged and unprivileged containers.

If someone cares about the bits that only work in privileged containers, they should just use the lxc-sabayon template directly.

geaaru:
I check lxc-sabayon script and when -u option is been set it is needed also --mapped-uid and --mapped-gid used by fluidshift for set right permissions.
Could be this a problem on automatic deploy system ?

stgraber:
Yes, that'd be a problem. I'm also rather confused as to why you even need to call fuidshift in there in the first place.

lxc-create runs your template script in a pre-mapped environment so if you create a file as uid 0 in there, it will be created as uid 100000 or whatever on the filesystem.

So the whole unprivileged_shift_owner function should be entirely unnecessary.

geaaru:
I just now execute lxc-create for an unprivileged container without unprivileged_shift_owner function on template script but at end of lxc-create script I have rootfs without shifted uids.
And so on lxc-start I receive:

lxc-start: conf.c: mount_rootfs: 807 Permission denied - failed to get real path for '/data/lxc/sab-unpriv3/rootfs'

So, if I remove unprivileged_shift_owner I can't use than lxc-sabayon script directly with lxc-create command.

Currently, I use this lxc version:

``
driver: lxc
driverversion: 2.0.6
kernel: Linux
kernelarchitecture: armv7l
kernelversion: 4.9.8-geaaru
server: lxd
serverpid: 12906
serverversion: "2.8"
storage: btrfs
storageversion: "4.9"

``

Could be a solution execute fluidshift command only if is present for example with another option like  --exec-fluidshift ? So, on lxc-ci I leave only -u option?

stgraber:
None of the other templates supporting unprivileged containers are doing any of the mapping, so the problem here is the sabayon template or your container's config.

stgraber:
If /rootfs is all owned by the real root, it suggests that the config which was provided to lxc-create (the default is at ~/.config/lxc/default.conf but can be overriden with -f) doesn't contain a uid/gid map.

geaaru:
But all others templates doesn't support directly creation of unprivileged containers. All others templates return a message like this:

```
    if [ "$arg" = "--mapped-uid" -o "$arg" = "--mapped-gid" ]; then
        echo "This template can't be used for unprivileged containers." 1>&2
        echo "You may want to try the \"download\" template instead." 1>&2
        exit 1
```

None of the other templates permit directly creation of unprivileged containers. 
I think that  have this feature could be fine.
However, I disable on default use of _unprivileged_shift_owner function so it is possible create image to use with lxc-download script like others template and for users that want create directly unprivileged containers I add an optional argument.

geaaru:
With this pull request https://github.com/lxc/lxc/pull/1465 probably this patch works fine.
Can you try to build images on jenkins before merge pull request for check that all works fine?

Thanks

stgraber:
create_message still needs changing to stop saying unprivileged.

geaaru:
I don't understand. Option -u  is always needed, but it is a static option. As you said me from your side jenkins build script will create an unique image for both privileged and unprivileged containers so it is needed that function unprivileged_rootfs must be always executed. Now it is possible use of -u option without --mapped-gid and --mapped-uid to permit creation of a rootfs compliant with both containers type without execute fuidshift.

stgraber:
@geaaru please read my comment. I was talking about "create_message", not about the arguments.

stgraber:
I want the ", u=Unprivileged Container" string gone from the create message as that doesn't mean anything when using lxc-download.

geaaru:
ah ok. Sorry. Done.
Thank you for support.

geaaru:
Hi, last question... I see jenkins jobs but I don't find any elaboration for arm architecture of lxc-sabayon template (only amd64) . Is needed some other activity ?
Thanks in advance

stgraber:
I've added armhf to Jenkins now

PR #22, opened 2018-04-01
alpine linux: add arm64 arch

AlbanVidal:
Please add arm64 arch for alpine linux
Signed-off-by: Alban Vidal alban.vidal@zordhak.fr


AlbanVidal:
jenkins: test this please

PR #42, opened 2019-06-24
images: Fix Ubuntu Core 18 init

monstermunchkin:


monstermunchkin:
It turns out that Core 18 has a `snapd.snap` as well which still needs to be handled.

PR #43, opened 2019-07-01
images: Do not create TTYs on Gentoo

monstermunchkin:


ste7677:
What's the reason behind not creating the default gentoo TTYs?

I ask because until this patch one could always easily connect to a gentoo container using "lxc-console -n containername" (assumed default behavior?). This behavior is now broken. Without specifying "-t 0" you can't connect anymore. Took me a long time to find out since I'm not an LXC expert. Anyway, I've added a note on the gentoo wiki LXC page.

PR #50, opened 2019-07-25
Add Void Linux

monstermunchkin:

### DIFF
--- a/images/voidlinux.yaml
+++ b/images/voidlinux.yaml
@@ -1,0 +1,132 @@
+image:
+  distribution: "voidlinux"
+
+source:
+  downloader: voidlinux-http
+  url: https://mirrors.servercentral.com/voidlinux/live/current/
+  keys:
+    # 0xCF24B9C038097D8A44958E2C8DEBDA68B48282A4
+    - |-
+      -----BEGIN PGP PUBLIC KEY BLOCK-----
+
+      mQENBFmFVJkBCACzAXGvRdC4rBfgOHGPmk7+apjnhhO8GYX/igc2Qk4jn68TewJA
+      Dxw3VFfn9yBzbV27WuFypN98D2Gw6DE+FXmxXqCx8ERj41A97VYWT0gQedWlIS+Y
+      QzawbLcpK0Ei+LABL33eDff+IaeHS8GgFqvfEaQMU2pf9oMFxQtAM5zfgHRUD5bM
+      5TEHnY4sRVa8Bm9VbzJKbyiQkI7Kzz5bVM/mbgQrwZL14BixytzR1XfuGBmDOC5D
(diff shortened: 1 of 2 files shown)


images/void.yaml
```diff
@@ -0,0 +1,132 @@
...
+targets:
+  lxc:
+    create-message: |-
+      You just created a {{ image.description }} container.
+
+    config:
+      - type: all
+        before: 5
+        content: |-
+          lxc.include = LXC_TEMPLATE_CONFIG/void.common.conf
```
stgraber: needs to be `voidlinux.common.conf` to match LXC's configs.

images/void.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+      - type: user
+        before: 5
+        content: |-
+          lxc.include = LXC_TEMPLATE_CONFIG/void.userns.conf
```
stgraber: needs to be `voidlinux.userns.conf` to match LXC's configs.

jenkins/jobs/image-void.yaml
```diff
@@ -0,0 +1,92 @@
+- job:
+    name: "image-void"
```
stgraber: I think we want this to be image-voidlinux (and rename the .yaml files too) so we match the name of the old LXC template. It also looks like this is the generally used name on mirrors and other places for that distro.

jenkins/jobs/image-void.yaml
```diff
@@ -0,0 +1,92 @@
...
+        LXD_ARCHITECTURE="${architecture}"
+
+        ARCH=${architecture}
+        [ "${ARCH}" = "amd64" ] && ARCH="x86_64"
+        [ "${ARCH}" = "armel" ] && ARCH="armv7l"
+        [ "${ARCH}" = "arm64" ] && ARCH="aarch64"
+        [ "${ARCH}" = "i386" ] && ARCH="i686"
+
+        if [ "${variant}" != "default" ]; then
+          exec sudo /lxc-ci/bin/build-distro /lxc-ci/images/void.yaml \
```
stgraber: Will need to rename to voidlinux

jenkins/jobs/image-void.yaml
```diff
@@ -0,0 +1,92 @@
...
+              ${LXD_ARCHITECTURE} 1800 ${WORKSPACE} \
+              -o image.architecture=${ARCH} \
+              -o image.release=${release} \
+              -o image.variant=${variant} \
+              -o source.variant=${variant} \
+              -o source.url="https://mirrors.servercentral.com/voidlinux/live/current/
+        else
+          exec sudo /lxc-ci/bin/build-distro /lxc-ci/images/void.yaml \
```
stgraber: Will need to rename to voidlinux

PR #54, opened 2019-08-08
Add cloud images

monstermunchkin:


stgraber:
How difficult would be to retain the single yaml per distro and make it possible to have the needed files, packages and script logic be conditional on the variant?

It feels like this would make it easier to introduce and maintain variants.

monstermunchkin:
This should work https://github.com/lxc/distrobuilder/pull/211

stgraber:
@monstermunchkin can you update this to use the existing .yaml files and just have them do the right thing when asked to build for the `cloud` variant?

PR #55, opened 2019-08-09
Port remaining jenkins jobs

monstermunchkin:


jenkins/jobs/lxc-build-android.yaml
```diff
@@ -0,0 +1,30 @@
...
+    builders:
+      - shell: |-
+          cd /lxc
+          exec sudo /lxc/bin/build-android amd64 https://github.com/lxc/lxc master ${sha1} ${WORKSPACE}/
+
+    publishers:
+      - archive:
+          artifacts: lxc-android.tar.gz
+
+      - workspace-cleanup
```
stgraber: Our standard pattern for this is to do:
```
- workspace-cleanup:
  fail-build: false
```

Can you update them all?

jenkins/jobs/lxc-build-android.yaml
```diff
@@ -0,0 +1,30 @@
...
+    parameters:
+      - string:
+          name: sha1
+          description: Hash of the commit to build
+
+      - string:
+          name: branch
+          description: Source git branch
+
+    builders:
```
stgraber: Small nit but in the existing jobs, this goes to the top, right after project-type and before properties.

stgraber:
Can you also make the yaml look more similar to our existing jobs so we can diff them without hitting issues due to differing indent?

And it looks like we have all our current jobs running ansicolor so would be nice for the newly ported ones to also use that.

Thanks!

jenkins/jobs/lxc-build-android.yaml
monstermunchkin: Done

jenkins/jobs/lxc-build-android.yaml
monstermunchkin: Done

monstermunchkin:
All done

PR #64, opened 2019-08-20
jenkins: Disable some Gentoo cloud images

monstermunchkin:
Cloud-init is only supported on amd64 and x86 [0].

[0] https://packages.gentoo.org/packages/app-emulation/cloud-init


stgraber:
Wouldn't something like:

`!(architecture!="amd64" && architecture!="i386" && variant=="cloud")`

Be better based on your description of the restriction?

monstermunchkin:
Fixed

PR #76, opened 2019-09-10
Fix cloud-init issues

monstermunchkin:


monstermunchkin:
I'm working on a fix for openSUSE Tumbleweed. `cloud-init` doesn't detect Tumbleweed properly since it explicitly looks for the string `opensuse` not `opensuse-tumbleweed`.

PR #81, opened 2019-09-21
Configure the hostname properly for gentoo hosts in /etc/conf.d/hostname.

vStone:
Having invalid configuration syntax in /etc/conf.d breaks the initial network setup in cloud-init.


vStone:
Looks like this is not quite right.

PR #85, opened 2019-10-01
images/debian: Also mask systemd-networkd-wait-online

sayap:
After fixing the shutdown issue (#4939) by masking systemd-networkd
(c63ceec4), we should also mask the systemd-networkd-wait-online
service, otherwise things that depend on network-online.target may hang
for 2 minutes until the service times out.

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -904,6 +904,7 @@ actions:
       # Disable networkd (unused)
       systemctl mask systemd-networkd.service
       systemctl mask systemd-networkd.socket
+      systemctl mask systemd-networkd-wait-online.service
 
       # Make sure the locale is built and functional
       echo en_US.UTF-8 UTF-8 >> /etc/locale.gen


sayap:
Steps to reproduce:

1. lxc launch images:debian/9 xxx
2. lxc exec xxx /bin/bash
3. create a dummy init script that depends on $network:
```
cat > /tmp/test <<'EOF'
> #!/lib/init/init-d-script
> ### BEGIN INIT INFO
> # Provides:   test
> # Required-Start:    $syslog $time $remote_fs
> # Required-Stop:     $syslog $time $remote_fs
> # Should-Start:      $network
> # Should-Stop:       $network
> # Default-Start:     2 3 4 5
> # Default-Stop:      0 1 6
> # Short-Description: test init script
> # Description:Debian init script to test debian init system
> ### END INIT INFO
> DAEMON=/usr/bin/test
> EOF
```
4. start the dummy service with:
```
chmod a+x /etc/init.d/test
update-rc.d test defaults
invoke-rc.d test start
```
Without masking systemd-networkd-wait-online, the last step hangs for 2 minutes before proceeding.
After masking systemd-networkd-wait-online, the last step succeeds immediately.

Found this issue while trying to install percona-xtradb-cluster

PR #106, opened 2020-01-25
Fully enable cloud-init on Fedora

mikemccllstr:
This change enables the Config and Final boot stages of cloud-init on the cloud variant of the Fedora image. This is necessary for all the documented features of cloud-init to work, such as the "packages" and "runcmd" config keys.

Signed-off-by: Michael McCallister <mike@mccllstr.com>


stgraber:
Thanks!

PR #108, opened 2020-02-20
images/openwrt: Ensure /var/lock

monstermunchkin:
This ensures that /var/lock is always present, especially when running
distrobuilder pack-lx{c,d}.
Due to /var being a symlink to /tmp which in turn is a tmpfs, the
directory /var/lock will never be persistent.

### DIFF
--- a/images/openwrt.yaml
+++ b/images/openwrt.yaml
@@ -403,6 +403,20 @@ actions:
       mkdir -p /var/lock
       echo "console::askfirst:/usr/libexec/login.sh" >> /etc/inittab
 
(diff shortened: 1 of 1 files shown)


monstermunchkin:
@stgraber Alternatively, we could make sure that `opkg` creates the `/var/lock` directory if not present. It's a simple fix either way.

stgraber:
@monstermunchkin feels cleaner if done in distrobuilder, less yaml stuff to keep around for everyone.

stgraber:
Closing and let's do this in distrobuilder instead.

PR #112, opened 2020-02-26
images: Add Debian VM support

monstermunchkin:

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -830,7 +830,7 @@ files:
    generator: remove
 
  - path: /etc/machine-id
-   generator: remove
+   generator: dump
 
  - path: /etc/network/interfaces
    generator: dump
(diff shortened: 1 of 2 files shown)


stgraber:
Missing the Jenkins part

monstermunchkin:
Since I have only tested it on amd64, I didn't include any other architectures. Is that fine for now?

stgraber:
You should be able to add arm64 without too much hassle, the rest are much trickier and I haven't done it for Ubuntu either.

Looking at the delta on ubuntu.yaml, we're missing:
 - machine-id should be created as an empty file, not be removed
 - network config is wrong and needs to use enp5s0 on vm
 - linux-image-amd64 needs to be put in an amd64-only section
 - linux-image-arm64 needs to be used for arm64
 - I don't believe `shim-signed` exists on all Debian releases, it should only be installed when available
 - Are you sure we don't need the grub sed workaround when building this image inside a container?

`git diff cb4cef97c0583e978d8519533a331f7063618c6d.. images/ubuntu.yaml ` is what I used to compare this against.

images/debian.yaml
```diff
@@ -831,6 +831,13 @@ files:
 
  - path: /etc/machine-id
    generator: remove
+   types:
+    - container
+
+ - path: /etc/machine-id
+   generator: dump
+   types:
+    - container
```
stgraber: You probably meant `vm` there instead of `container` and we in fact just want it to be empty in both cases, same way we've done it in ubuntu.

images/debian.yaml
```diff
@@ -867,6 +910,15 @@ files:
...
 
+ - name: lxd-agent
+   generator: lxd-agent
+   types:
+     - vm
+
+ - generator: fstab
+   types:
+    - vm
+
```
stgraber: Indent doesn't match between those two.

stgraber:
Rest looks fine, so with those two things fixed, we can merge.

images/debian.yaml
```diff
@@ -831,6 +831,13 @@ files:
 
  - path: /etc/machine-id
    generator: remove
+   types:
+    - container
+
+ - path: /etc/machine-id
+   generator: dump
+   types:
+    - vm
```
stgraber: Why the differing logic?

monstermunchkin: Misread your previous comment. Fixed it.

PR #115, opened 2020-03-02
Add Fedora VM support

monstermunchkin:


images/fedora.yaml
```diff
@@ -36,6 +36,42 @@ targets:
...
+      - vm
+
+  - generator: fstab
+    types:
+      - vm
+
+  - path: /etc/machine-id
+    generator: dump
+    types:
+      - vm
```
stgraber: Any reason not to do it for containers too, same as we do for Debian/Ubuntu?

images/fedora.yaml
```diff
@@ -140,6 +180,16 @@ packages:
       - sssd-client
      action: remove
 
+   - packages:
+     - grub2-efi-x64
+     - kernel
+     - shim
+     - btrfs-progs
```
stgraber: Hmm, why btrfs?

images/fedora.yaml
stgraber: Actually, given you're now always generating a machine-id as part of the image, we absolutely need to clear the file in both containers and vms, or we'll end up with all our containers having the same unique id.

images/fedora.yaml
monstermunchkin: IIRC something complained about btrfs missing.

images/fedora.yaml
monstermunchkin: Yeah, that's wrong. I'll fix that.

images/fedora.yaml
monstermunchkin: I'll test this again.

monstermunchkin: I removed it. I had a reason for adding it but I don't remember why. Anyway, it works fine without it.

PR #133, opened 2020-03-10
Add Alpine VM support

monstermunchkin:


images/alpine.yaml
```diff
@@ -195,6 +195,35 @@ files:
...
+  content: |-
+    # Set the recordfail timeout
+    GRUB_RECORDFAIL_TIMEOUT=0
+
+    # Do not wait on grub prompt
+    GRUB_TIMEOUT=0
+
+    # Set the default commandline
+    # FIXME: Don't hard-code filesystem
+    GRUB_CMDLINE_LINUX_DEFAULT="console=tty1 console=ttyS0 modules=sd-mod,usb-storage,ext4 rootfstype=ext4"
```
monstermunchkin: We need a way to not hard-code the filesystem. This could be done by using pongo in the dump generator.

images/alpine.yaml
```diff
@@ -241,6 +270,26 @@ files:
...
+    #!/sbin/openrc-run
+
+    description="LXD - agent - 9p mount"
+    command=/bin/mount
+    command_args="-t 9p config /run/lxd_config/9p -o access=0,trans=virtio"
+
+    depend() {
+      want udev-trigger
+      after udev-trigger
+    }
```
monstermunchkin: This might be specific to Alpine. The only other distro using openRC is Gentoo which I will look at next. If Gentoo needs this as well, I will add it to the lxd-agent generator.

monstermunchkin:
This is blocked by https://github.com/lxc/distrobuilder/pull/336.

images/alpine.yaml
monstermunchkin: See https://github.com/lxc/distrobuilder/pull/336

images/alpine.yaml
stgraber: We have that generator now so we should use it here.

PR #134, opened 2020-03-10
Add Gentoo VM support

monstermunchkin:


monstermunchkin:
This is still WIP. Some hard-coded amd64 code needs to be fixed. And `lxd-agent` is panicking with
```
panic: Unix syslog delivery error
...
```

which seems to be caused by a missing syslogger (although `syslog-ng` is installed).

stgraber:
Ah, maybe we can do something a bit more clever in `lxd-agent`. We certainly don't want to crash on missing syslog. We may just want to spit out everything on stdout/stderr an let systemd or whatever the init system is handle it.

images/gentoo.yaml
```diff
@@ -411,6 +444,8 @@ actions:
     cp /usr/share/portage/config/repos.conf /usr/share/portage/config/repos.conf.orig
     sed -i "s#hkps://keys.gentoo.org#keyserver.ubuntu.com#g" /usr/share/portage/config/repos.conf
     sed -ri "s#(sync-uri =) .+#\1 ${RSYNC_MIRROR}#" /usr/share/portage/config/repos.conf
+
+    echo ACCEPT_KEYWORDS="~amd64" >> /etc/portage/make.conf
```
stgraber: hmm, that will break other architectures

images/gentoo.yaml
```diff
@@ -452,6 +487,15 @@ actions:
...
   - cloud
 
+- trigger: post-files
+  action: |-
+    #!/bin/sh
+    set -eux
+    grub-install --target=x86_64-efi --no-nvram --removable
+    grub-install --target=x86_64-efi --no-nvram
+    grub-mkconfig -o /boot/grub/grub.cfg
+    sed -i "s#root=[^ ]*#root=/dev/sda2#g" /boot/grub/grub.cfg
```
stgraber: This would also need to be arch-specific.

monstermunchkin:
Fixed the syslog thing, and the hard-coded architecture code as well. With https://github.com/lxc/distrobuilder/pull/307, this gives us a working Gentoo VM.

monstermunchkin:
For some reason the VM cannot be stopped using `lxc stop`. I think we had similar issues with Gentoo containers in the past. That's still something which needs fixing here.

stgraber:
For the container it was probably just bad signal handling but for a VM, it's probably ACPI related.

Check if the init system can handle acpi events directly, if not you may need acpid or something.

monstermunchkin:
Installing acpid did it, thanks! Now it should be complete.

PR #135, opened 2020-03-13
Add PLD Linux

mistoo:
PLD (https://pld-linux.org/) LXD/LXC image. Tested with bin/build-distro script, but I'm not 100% sure about Jenkins job.

### DIFF
--- a/images/pld.yaml
+++ b/images/pld.yaml
@@ -1,0 +1,146 @@
+image:
+  distribution: pld
+
+source:
+  downloader: docker-http
+  url: pldlinux/i686:latest
+
+targets:
+  lxc:
+    create-message: |-
+      You just created a {{ image.description }} container.
+    config:
+    - type: all
+      content: |-
+        lxc.include = LXC_TEMPLATE_CONFIG/common.conf
+    - type: user
+      content: |-
+        lxc.include = LXC_TEMPLATE_CONFIG/userns.conf
+    - type: all
+      content: |-
+        lxc.arch = {{ image.architecture_personality }}
+
+packages:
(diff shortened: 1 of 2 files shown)


images/pld.yaml
```diff
@@ -0,0 +1,146 @@
+image:
+  distribution: pld
+
+source:
+  downloader: docker-http
+  url: pldlinux/i686:latest
```
stgraber: The `/i686` here would indicate that you're only ever pulling the 32bit image.
If there's a 64bit image, then hopefully just omitting the architecture will do the right thing.

If there's only a 32bit image available, then you shouldn't list amd64 as an architecture in the jenkins job.

stgraber:
We're going to have to test it before merging as I'm not sure that Docker can be used in our image build environment. It would be the first image making use of this, so this may be a problem.

images/pld.yaml
mistoo: Yep, this is placeholder actually, as source.url is replaced by -o in jenkins job; as far as I know  distrobuilder is not able to handle template inlining here ({{ architecture }})

stgraber: Ah, right, in this case it's probably best not to set it at all so we don't get confused again when reading it :)

mistoo: Other readers may have no clue what is the image source then, maybe I just comment out this?

stgraber: Yeah, I guess you can comment it out.

For other images, they're almost all lacking the `image.release` property, so folks using the YAMLs without knowing to feed the missing stuff will have issues already :)

PR #138, opened 2020-03-24
images: Add acpid to VM images

monstermunchkin:
This adds acpid to Debian and Ubuntu VM images which resolves shutdown
issues.


stgraber:
@monstermunchkin this should apply to both amd64 and arm64

PR #143, opened 2020-04-16
Revert "images: Remove 10-lxc.yaml for cloud images"

monstermunchkin:


stgraber:
This gets us back to having two different config files in netplan, conflicting with each other, so that's not the way to go.

PR #144, opened 2020-04-16
Enable systemd-networkd on Ubuntu/Mint cloud images

monstermunchkin:


images/mint.yaml
```diff
@@ -407,6 +407,17 @@ actions:
     getent group sudo >/dev/null 2>&1 || groupadd --system sudo
     useradd --create-home -s /bin/bash -G sudo -U ubuntu
 
+- trigger: post-packages
+  action: |-
+    #!/bin/sh
+    set -eux
+
+    # Enable systemd-networkd
```
stgraber, line 415: Any reason not to just run `systemctl enable systemd-networkd`?
Because based on my local tests, you're missing a bunch of links here.

You also need this restricted to those releases that use netplan but not restricted to cloud.
I don't know by what magic things are working for non-cloud right now but we may as well be consistent.

monstermunchkin: Should be fine now

PR #152, opened 2020-05-11
images/ubuntu: Fix file mode on sudoers file

monstermunchkin:


stgraber:
I'd like this to be done through YAML only not using an action

PR #164, opened 2020-05-26
images/sabayon: Fix IP issue

monstermunchkin:


stgraber:
I'm confused why can't this use the same fix as the cloud variant?

monstermunchkin:
I was confused. You're absolutely right :)

Issue #166, opened 2020-06-10
Cloud-init network configuration fails in Debian/buster/cloud/amd64 image 


mjrider:
Enviroment:
network: without dhcp, all interfaces need a static allocation

partial config with both methods of defining the dns resolvers
```
architecture: x86_64
config:
  boot.autostart: "1"
  image.architecture: amd64
  image.description: Debian buster amd64 (20200609_05:24)
  image.os: Debian
  image.release: buster
  image.serial: "20200609_05:24"
  user.network-config: "version: 1\nconfig:\n  - type: physical\n    name: eth0\n
    \   subnets:\n      - type: static\n        address: 10.53.2.170\n        netmask:
    255.255.255.0\n        gateway: 10.53.2.1\n        control: auto\n  - type: nameserver\n
    \   address: \n      - 10.53.1.101\n      - 10.53.1.102\n"
  user.user-data: |
    #cloud-config
    manage_etc_hosts: true
    manage_resolv_conf: true

    resolv_conf:
      nameservers:
        - 10.53.1.101
        - 10.53.1.102
      options:
        rotate: true
        timeout: 1
  volatile.base_image: 54d80ea17c872cc0964c55a219c1625375f313d25619aabd9e05936a5b1a2120

```

This results in:
- empty /etc/resolv.conf file, because the package resolvconf ( or alternative ) is missing
- slow boot, because of hardcoded eth0 dhcp in /etc/network/interfaces (https://github.com/lxc/lxc-ci/blob/master/images/debian.yaml#L856)

PR #167, opened 2020-06-11
Fixing cloud-init issue with network and dns resolving setup

mjrider:
Hopefully this is the right fix for #166 
it works for me for debian buster with cloud-init.

points of attention:
- i picked openresolv instead of resolvconf because resolvconf fails to install during the distrobuild because /etc/resolv.conf isn't an file

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -859,6 +859,24 @@ files:
     source /etc/network/interfaces.d/*.cfg
   types:
   - container
+  variants:
+  - default
+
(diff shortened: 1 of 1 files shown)


mjrider:
tested this manual for amd64
- buster, variant default
- buster, variant cloud
- stretch, variant cloud

stgraber:
Thanks. We'll see if anyone complains, if they do, we'll sort out resolvconf as that's what I believe we have in most other images.

Issue #184, opened 2020-08-01
gentoo image has empty repository path /var/db/repos/gentoo


pajot:
The base Gentoo image (966106a9d0cd) is missing the /var/db/repos/gentoo directory, which means that portage cannot be synced and the system is not modifiable by conventional means:

```
gentooC01 # emerge -av app-portage/eix
!!! Section 'gentoo' in repos.conf has location attribute set to nonexistent directory: '/var/db/repos/gentoo'
!!! Invalid Repository Location (not a dir): '/var/db/repos/gentoo'


!!! /etc/portage/make.profile is not a symlink and will probably prevent most merges.
!!! It should point into a profile within /var/db/repos/gentoo/profiles/
!!! (You can safely ignore this message when syncing. It's harmless.)


!!! Your current profile is invalid. If you have just changed your profile
!!! configuration, you should revert back to the previous configuration.
!!! Allowed actions are limited to --help, --info, --search, --sync, and
!!! --version.
gentooC01 # ls /var/db/repos/
gentooC01 # eselect profile list
!!! Error: Failed to get a list of valid profiles
exiting
gentooC01 # emerge --info
!!! Section 'gentoo' in repos.conf has location attribute set to nonexistent directory: '/var/db/repos/gentoo'
!!! Invalid Repository Location (not a dir): '/var/db/repos/gentoo'
WARNING: One or more repositories have missing repo_name entries:

        /var/db/repos/gentoo/profiles/repo_name

NOTE: Each repo_name entry should be a plain text file containing a
unique name for the repository on the first line.


Portage 2.3.103 (python 3.7.8-final-0, unavailable, gcc-9.3.0, glibc-2.30-r8, 4.19.97-gentoo x86_64)
=================================================================
System uname: Linux-4.19.97-gentoo-x86_64-Intel-R-_Core-TM-_i7-7600U_CPU_@_2.80GHz-with-gentoo-2.6
KiB Mem:    32804508 total,  23478692 free
KiB Swap:   32813020 total,  32813020 free
sh bash 5.0_p17
ld GNU ld (Gentoo 2.33.1 p2) 2.33.1
dev-lang/python:          2.7.18-r1::gentoo, 3.7.8-r2::gentoo, 3.8.4-r1::gentoo
sys-devel/autoconf:       2.69-r4::gentoo
sys-devel/automake:       1.16.1-r1::gentoo
sys-devel/binutils:       2.33.1-r1::gentoo
[...]

PR #185, opened 2020-08-02
Add init script to emerge --sync at first boot.

pajot:
Here is my attempt at solving the problem. I was unable to create a working image on my system, distrobuilder fails with errors.

I get this output:

```
# /usr/local/bin/distrobuilder build-lxd gentoo.yaml 
+ mv /usr/bin/gpg2 /usr/bin/gpg2.real
+ echo '#!/bin/sh'
+ echo 'exec /usr/bin/gpg2.real --keyserver keyserver.ubuntu.com "$@"'
+ chmod +x /usr/bin/gpg2
+ cp /usr/share/portage/config/repos.conf /usr/share/portage/config/repos.conf.orig
+ sed -i s#hkps://keys.gentoo.org#keyserver.ubuntu.com#g /usr/share/portage/config/repos.conf
+ sed -ri 's#(sync-uri =) .+#\1 rsync://rsync.ca.gentoo.org/gentoo-portage/#' /usr/share/portage/config/repos.conf
+ '[' -e /etc/inittab ']'
+ sed -i 's/^c[0-9]:/#\0/' /etc/inittab
+ sed -i 's/^#\(x1:.*\)/\1/' etc/inittab
+ echo pf:12345:powerwait:/sbin/halt
+ mv /usr/bin/gpg2 /usr/bin/gpg2.real
+ echo '#!/bin/sh'
+ echo 'exec /usr/bin/gpg2.real --keyserver keyserver.ubuntu.com "$@"'
+ chmod +x /usr/bin/gpg2
+ cp /usr/share/portage/config/repos.conf /usr/share/portage/config/repos.conf.orig
+ sed -i s#hkps://keys.gentoo.org#keyserver.ubuntu.com#g /usr/share/portage/config/repos.conf
+ sed -ri 's#(sync-uri =) .+#\1 rsync://rsync.ca.gentoo.org/gentoo-portage/#' /usr/share/portage/config/repos.conf
++ uname -m
+ TARGET=x86_64
+ case "${TARGET}" in
+ TARGET=amd64
+ echo 'ACCEPT_KEYWORDS=~amd64'
+ ln -s net.lo /etc/init.d/net.eth0
+ mkdir -p /etc/runlevels/default
+ rc-update add net.eth0 default
 * service net.eth0 added to runlevel default
+ rm /usr/bin/gpg2
+ mv /usr/bin/gpg2.real /usr/bin/gpg2
+ rm /usr/share/portage/config/repos.conf
+ mv /usr/share/portage/config/repos.conf.orig /usr/share/portage/config/repos.conf
+ rm /usr/bin/gpg2
+ mv /usr/bin/gpg2.real /usr/bin/gpg2
mv: cannot stat '/usr/bin/gpg2.real': No such file or directory
Error: Failed to run post-packages: exit status 1
```

It looks like distrobuilder is executing both post-packages sections:
https://github.com/lxc/lxc-ci/blob/5865c94b68256463d2bceaf32c9437d1cf059aa1/images/gentoo.yaml#L469-L493

If I comment out the second one, I get this output:
```
# /usr/local/bin/distrobuilder build-lxd gentoo.yaml 
+ mv /usr/bin/gpg2 /usr/bin/gpg2.real
+ echo '#!/bin/sh'
+ echo 'exec /usr/bin/gpg2.real --keyserver keyserver.ubuntu.com "$@"'
+ chmod +x /usr/bin/gpg2
+ cp /usr/share/portage/config/repos.conf /usr/share/portage/config/repos.conf.orig
+ sed -i s#hkps://keys.gentoo.org#keyserver.ubuntu.com#g /usr/share/portage/config/repos.conf
+ sed -ri 's#(sync-uri =) .+#\1 rsync://rsync.ca.gentoo.org/gentoo-portage/#' /usr/share/portage/config/repos.conf
+ '[' -e /etc/inittab ']'
+ sed -i 's/^c[0-9]:/#\0/' /etc/inittab
+ sed -i 's/^#\(x1:.*\)/\1/' etc/inittab
+ echo pf:12345:powerwait:/sbin/halt
+ mv /usr/bin/gpg2 /usr/bin/gpg2.real
+ echo '#!/bin/sh'
+ echo 'exec /usr/bin/gpg2.real --keyserver keyserver.ubuntu.com "$@"'
+ chmod +x /usr/bin/gpg2
+ cp /usr/share/portage/config/repos.conf /usr/share/portage/config/repos.conf.orig
+ sed -i s#hkps://keys.gentoo.org#keyserver.ubuntu.com#g /usr/share/portage/config/repos.conf
+ sed -ri 's#(sync-uri =) .+#\1 rsync://rsync.ca.gentoo.org/gentoo-portage/#' /usr/share/portage/config/repos.conf
++ uname -m
+ TARGET=x86_64
+ case "${TARGET}" in
+ TARGET=amd64
+ echo 'ACCEPT_KEYWORDS=~amd64'
+ ln -s net.lo /etc/init.d/net.eth0
+ mkdir -p /etc/runlevels/default
+ rc-update add net.eth0 default
 * service net.eth0 added to runlevel default
+ rm /usr/bin/gpg2
+ mv /usr/bin/gpg2.real /usr/bin/gpg2
+ rm /usr/share/portage/config/repos.conf
+ mv /usr/share/portage/config/repos.conf.orig /usr/share/portage/config/repos.conf
Error: Failed to create overlay: no such device
```

### DIFF
--- a/images/gentoo.yaml
+++ b/images/gentoo.yaml
@@ -343,6 +343,14 @@ files:
     rc_keyword="-stop"
     config_eth0="dhcp"
 
+- path: /etc/local.d/repo_check.start
+  generator: dump
+  mode: 700
+  content: |-
+    #!/bin/bash
+    filename="header.txt"
+    test -f $(portageq get_repo_path / gentoo)/$filename || emerge --sync; chmod -x /etc/local.d/repo_check.start
+
 - name: meta-data
   generator: cloud-init
   variants:


stgraber:
You'll need to include a `Signed-off-by: Firstname Lastname <email>` line at the end of your commit message.
`git commit --amend -s && git push --force` can do that for you usually.

Is that local script guaranteed to run after the network init scripts or is there some more logic to put in place to have this kick in once network is up?

As for building it, try passing `-o image.architecture=amd64 -o image.variant=container` that may fix your issues.

monstermunchkin:
>  I was unable to create a working image on my system, distrobuilder fails with errors.

That's a bug in the current gentoo.yaml. Processing both triggers is correct but we need to make sure, the files we are moving around actually exist.

monstermunchkin:
@pajot My gentoo image/container is missing the header.txt and therefore the script doesn't run on first boot.

pajot:
> @pajot My gentoo image/container is missing the header.txt and therefore the script doesn't run on first boot.

@monstermunchkin I'm sorry, my mistake. I've force pushed the fix. Does this work?

(It would help to fix gentoo.yaml so that I can do local tests before pushing my fixes :) )

pajot:
> > I was unable to create a working image on my system, distrobuilder fails with errors.
> 
> That's a bug in the current gentoo.yaml. Processing both triggers is correct but we need to make sure, the files we are moving around actually exist.

@monstermunchkin There must be multiple bugs in gentoo.yaml; when I comment out the second trigger distrobuilder fails with 'Error: Failed to create overlay: no such device', see above.

monstermunchkin:
> There must be multiple bugs in gentoo.yaml; when I comment out the second trigger distrobuilder fails with 'Error: Failed to create overlay: no such device', see above.

@pajot What version of distrobuilder are you using? The master branch?

pajot:
> > There must be multiple bugs in gentoo.yaml; when I comment out the second trigger distrobuilder fails with 'Error: Failed to create overlay: no such device', see above.
> 
> @pajot What version of distrobuilder are you using? The master branch?

I believe so, I followed the build instructions in the readme and built it with/within go.

EarnyHerzberg:
Just landed on the banning list of gentoo while creating a bunch of containers created by a buggy ansible scipt.
Commit ad30bdd is not a good idea, the script will run in the background before you have a chance to setup local mirrors.
Even worse it iteract badly with scripts thad doing the same, you don't know when the script has finished.

Done basically with lxd:
lxc launch <gentoo>

stgraber:
@monstermunchkin

pajot:
@EarnyHerzberg So, how would go about fixing this? Not having a populated main tree leaves you with a crippled container, so just not doing this is not a solution.

pajot:
@EarnyHerzberg 
> Just landed on the banning list of gentoo while creating a bunch of containers created by a buggy ansible scipt.
> Commit [ad30bdd](https://github.com/lxc/lxc-ci/commit/ad30bddb2264f39db271c48b228d03bfa10139ef) is not a good idea, the script will run in the background before you have a chance to setup local mirrors.
> Even worse it iteract badly with scripts thad doing the same, you don't know when the script has finished.

Can you elaborate on this? I don't see how this will interact badly with scripts. It runs once, and only if there is no main tree.

The point of this approach is to keep the image size small.

It takes quite a lot of abuse to get banned. How many containers did you launch at once?

EarnyHerzberg:
It's ok to start with an empty portage tree to save space. But for the first sync you may need some configuration first, in my case a local (fast!) mirror. 

My workaround at the moment are to create the container in a stopped state, delete the script, then start it up, reconfigure make.conf with a proxy and point repos.conf to a local mirror. After that run a sync, install git and switch repos.conf to use a local git mirror.

With the script in local.d this does not work. I've created 6 (then 4, 2) container at once, and was not aware about the early sync. My scripts where running, and one of the steps are delete the portage tree to switch to git. This failed miserably due to the fact that all containers still downloading the early sync i'm not aware of. So i tried to fix my scripts (why the f***k rm-rf fails?? :-) until i found a logmessage that i'm landed on the banning list.

The i was thinking about how to fix that with the early sync in place. And that is hard to impossible. After the first sync you have to select a profile for example, and this only works if the sync is finished. It's run in the 'background', sshd are already running ... if the sync failed, you don't see it. Or when it is finished? Without a mirror, slow internet connection and a lot of conatiners it can take a while.

pajot:
@EarnyHerzberg Yeah, multiple simultaneous syncs from a single IP is enough to get you banned. Sorry about that.

That just goes to show that the creativity of other people is always greater than you expect. I hadn't considered the possibility that someone might have a script that deletes the main tree and puts it elsewhere. Obviously that's not going to work with this local.d script.

I see two possible remedies:

1. build the main tree into the container image. It wouldn't be oppressively large, about 270 MB, if I recall correctly.
2. If that is still deemed too large, then an init script would need to be written, as suggested [on the Gentoo wiki page for local.d](https://wiki.gentoo.org/wiki//etc/local.d). This would at least block the boot process until the sync is complete and could be easily disabled with `rc-update`. This is best put in an ebuild and added to the Gentoo main tree. But that means it will have to be maintained as well. Perhaps it could be added to the lxc or lxd or distrobuilder ebuilds, which have maintainers.

Considering that, I think the easiest solution is 1. No magic required. Only disk space and a bit of bandwidth. (What do you think of that option, @stgraber?)

But a container that has no main tree at all to begin with is not user-friendly. It makes it impossible to update, add or remove packages.

EarnyHerzberg:
(don't  care about the banning, it does'nt last long, and if have mirrors, or use github for syncing)

You have to delete the portage tree if you want to sync with git. Steps are:
emerge --sync
emerge git
\<modify repos conf to use git\>
rm -rf /var/db/repos/gentoo/
emerge --sync

The reason is that git will not clone into a no-empty directory. And git is not part of the stage3 basesystem.

-->  1. sems easy, but is also a bad idea. The tree is outdated after a hour of image creation :-)
-->  2: Too complicated. Or: Create a container without initial internet connection ...

It should be enough to inform the user that he has to run 'emerge --sync' first. Even that is questionable: Then you should also display the gentoo handbook: Run 'eselect profile', change useflags, emerge -u @world, ... and so on ;-). Gentoo was not and is still not beginner friendly.

pajot:
> (don't care about the banning, it does'nt last long, and i have mirrors, or use github for syncing)
> 
> You have to delete the portage tree if you want to sync with git. Steps are:
> emerge --sync
> emerge git
> <modify repos conf to use git>
> rm -rf /var/db/repos/gentoo/
> emerge --sync
> 
> The reason is that git will not clone into a no-empty directory. And git is not part of the stage3 basesystem.
> 
> --> 1. sems easy, but is also a bad idea. The tree is outdated a hour after image creation :-)
> --> 2: Too complicated. Or: Create a container without initial internet connection ...

The problem is that without this script I was not able to sync inside the newly created container at all.

It doesn't matter how old the tree in the image is, it just has to let me sync.

Option 1 is increasingly looking like the only sensible one.

EarnyHerzberg:
The question is, why you can't sync, but the script can?

pajot:
> The question is, why you can't sync, but the script can?

My mistake. Syncing worked. Everything else didn't (due to the empty repo directory). The real history of this issue can be found in #184.

EarnyHerzberg:
Yea, i've just read it. User error :-) Now fixed with a script running at boot time. I agree that the script will help a beginner to make the first step with gentoo. But it hinders more experienced users. Every beginner will get more experienced over time (i hope). And then the script is still there....

pajot:
> Yea, i've just read it. User error :-) Now fixed with a script running at boot time. I agree that the script will help a beginner to make the first step with gentoo. But it hinders more experienced users. Every beginner will get more experienced over time (i hope). And then the script is still there....

I stand by my position that leaving the container without a repo after launching is not acceptable. I expect to have a working Gentoo install when I launch a container. Let's find a way to solve this.

pajot:
> Yea, i've just read it. User error :-) Now fixed with a script running at boot time. I agree that the script will help a beginner to make the first step with gentoo. But it hinders more experienced users. Every beginner will get more experienced over time (i hope). And then the script is still there....

@EarnyHerzberg How do you maintain your containers without a main tree? Perhaps there's an acceptable compromise that is practical but still keeps everything compact.

I considered adding an informative login message so that users aren't surprised when portage database queries fail.

EarnyHerzberg:
Yes. Good idea. Check DISTDIR, if empty, write a message...

PR #205, opened 2020-12-07
Adds test for LXD VMs

tomponline:
There's a few more tests to add, but this adds most of them:

Runs the following tests for dir, btrfs, lvm and zfs:

- VM create.
- VM grow on next reboot.
- VM shrink prevented.
- VM create after storage volume.size changed.
- VM create after storage volume.block.filesystem changed.
- VM create from profile with disk smaller than default VM volume size.
- VM migrate across pools with disk size larger than volume.size.
- VM migrate across pools with disk size smaller than volume.size.
- Export VM larger than default size and import (with/without volume.size set).

### DIFF
--- a/bin/test-lxd-vm
+++ b/bin/test-lxd-vm
@@ -1,0 +1,174 @@
+#!/bin/sh
+set -eux
+
+cleanup() {
+    echo ""
+    if [ "${FAIL}" = "1" ]; then
+        echo "Test failed"
+        exit 1
+    fi
+
+    echo "Test passed"
+    exit 0
+}
+
+poolDriverList="${1:-dir btrfs lvm zfs}"
+FAIL=1
+trap cleanup EXIT HUP INT TERM
+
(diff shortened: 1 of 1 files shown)


bin/test-lxd-vm
```diff
@@ -0,0 +1,152 @@
...
+    if [ "${FAIL}" = "1" ]; then
+        echo "Test failed"
+        exit 1
+    fi
+
+    echo "Test passed"
+    exit 0
+}
+
+poolDriverList="${1:-dir btrfs lvm zfs}"
```
stgraber, line 15: Any reason not to do ceph too?

```
curl -s http://canonical-lxd.stgraber.org/config/ceph.sh | sh
```

Should get you the credentials in place for it.

tomponline: @stgraber ah ok I'll provision lantea and try it with that.

PR #215, opened 2020-12-15
images/centos: Fix CentOS 8-Stream networking

monstermunchkin:


stgraber:
Hmm this feels like a very nasty workaround...

@monstermunchkin what's the issue with Network Manager that makes this kind of hack necessary?

monstermunchkin:
I don't know why NetworkManager is being weird. It behaved before. Anyway, there are two solutions (I know of) which fix the networking issue:
1. `ip link set dev eth0 up && nmcli c up "System eth0"`
2. Start the dhclient after NetworkManager

tomponline:
@stgraber @monstermunchkin I took a quick look at this an according to:

https://bugzilla.redhat.com/show_bug.cgi?id=1832603

If you comment out the line in /usr/lib/udev/rules.d/85-nm-unmanaged.rules that mentions veth, then it works without further modification.

tomponline:
According to that file we should be able to create an override file at `/etc/udev/rules.d`

monstermunchkin:
Let's try this. Thanks @tomponline :)

tomponline:
@monstermunchkin no problem. There does appear to be a separate issue, it looks like NM is preventing the container from cleanly shutting down, even with that fix (and before).

tomponline:
@monstermunchkin oh actually it does shutdown eventually its just slow.

PR #216, opened 2020-12-15
images/centos: Fix CentOS 8-Stream networking issue

monstermunchkin:


tomponline:
@monstermunchkin The top of that file says "# Do not modify this file, it will get overwritten on updates. # To override or extend the rules place a file in /etc/udev/rules.d" so it may be prudent to add a file in /etc/udev/rules.d to override the problem rule somehow.

tomponline:
Oh I see you've created a new file, although I wonder if using /etc/udev/rules.d would still be more appropriate. What do you think?

monstermunchkin:
You're right. I missed that.

tomponline (approved):
Cool, does that override fix it ok?

monstermunchkin:
Yes, the fix works.

PR #217, opened 2020-12-16
images/centos: Fix CentOS 8 VM

monstermunchkin:


monstermunchkin:
This fixes the networking issue which was caused by the VM not being able to boot.

PR #234, opened 2021-01-18
images/opensuse: improve cloud-init support

adamcstephens:
This fixes a few issues with the existing opensuse lxd images.

* Install `cloud-init-config-suse` explicitly. Currently `cloud-init-config-MicroOS` is being installed as a dependency to provide `/etc/cloud/cloud.cfg`, but this configuration disables some desirable modules including packages. The suse package seems to be a better choice for these images.
* Enable cloud-init config and final stages. These are disabled by default as well, and without them cloud-init loses a number of modules as configured.
* Write an empty fstab for cloud containers in order to silence the mounts module from warning.

### DIFF
--- a/images/opensuse.yaml
+++ b/images/opensuse.yaml
@@ -158,6 +158,14 @@ files:
   release:
   - 15.2
 
+- path: /etc/fstab
+  generator: dump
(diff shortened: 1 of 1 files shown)


stgraber:
Thanks!

PR #258, opened 2021-03-15
images/centos: Enable network-config for CentOS 7

monstermunchkin:


stgraber:
@monstermunchkin did you confirm that the cloud-init version in centos7 supports it?

stgraber:
Confirmed on forum that cloud-init got updated.

Issue #269, opened 2021-04-16
Feature request - add OpenWrt arm64 and armhf


ja-pa:
Hi,
would it be possible to add to OpenWrt jenkins job arm64(aarch64) and armhf(armv7) architectures ?
These images could be then used with lxc on arm based routers (see https://github.com/openwrt/packages/blob/master/utils/lxc/Makefile)

stgraber:
@monstermunchkin

gmiranda:
May I ask what's currently blocking this? I had a quick look at your build system and couldn't figure out how the hell it downloads any file from the OpenWrt page :rofl:

stgraber:
Not sure! @monstermunchkin

monstermunchkin:
@gmiranda The reason is that OpenWrt for ARM needs to be built on x86, but we don't support cross builds. You'll need to build the container yourself using `distrobuilder`.

stgraber:
Ah right, that was it, we don't support cross building on our infrastructure and running natively on ARM doesn't work.

gmiranda:
And I guess using OpenWRT's armvirt64/32 rootfs binary package is out ot the question, isn't it?

monstermunchkin:
I just looked into it.

We use https://github.com/mikma/lxd-openwrt as part of our build process. This should build a patches `procd`. However, I just discovered that none of the versions we build seem to need a patch. I'm therefore wondering if we need this repo at all.

Should this not be the case, we could simplify our build process and possibly provide armvirt images as well.

monstermunchkin:
https://github.com/lxc/distrobuilder/pull/542

gmiranda:
> I just looked into it.
> 
> We use https://github.com/mikma/lxd-openwrt as part of our build process. This should build a patches `procd`. However, I just discovered that none of the versions we build seem to need a patch. I'm therefore wondering if we need this repo at all.
> 
> Should this not be the case, we could simplify our build process and possibly provide armvirt images as well.

According to cvmiller/openwrt-lxd (and as far as I can see) there's no need to patch anything on versions 19.07 and upwards. I'll give that a try as soon as possible

monstermunchkin:
> > I just looked into it.
> > We use https://github.com/mikma/lxd-openwrt as part of our build process. This should build a patches `procd`. However, I just discovered that none of the versions we build seem to need a patch. I'm therefore wondering if we need this repo at all.
[...]

PR #270, opened 2021-04-19
Support openWrt ARM images

monstermunchkin:


monstermunchkin:
This closes #269

jenkins/jobs/image-openwrt.yaml
```diff
@@ -11,6 +11,8 @@
         type: slave
         values:
         - amd64
+        - aarch64
+        - armv7l
```
stgraber, line 15: Those aren't valid architectures, we'd expect `arm64` and `armhf` but that will cause the image being built on the native architecture which from what you've mentioned over on the distrobuilder issue won't actually work.

monstermunchkin: Should we then close this PR, and just have users build the images themselves with `distrobuilder`?

stgraber: We can merge the yaml change and just skip the jenkins change so they at least have a YAML they can use.

monstermunchkin: Done.

PR #282, opened 2021-05-15
debian: let cloud-init manage vm interfaces

adamcstephens:
The cloud variant of the vm type currently populates the dhcp interface config. This is detrimental for assigning static IPs with cloud-init, and duplicative of cloud-init's default of dhcp on the first interface.

This PR makes the vm type behavior match the existing container type, and sets no interface config for cloud variants.

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -862,22 +862,6 @@ files:
     auto lo
     iface lo inet loopback
 
-    source /etc/network/interfaces.d/*
-  types:
-  - container
-  variants:
-  - cloud
-
-- path: /etc/network/interfaces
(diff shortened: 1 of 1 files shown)


stgraber:
Looks reasonable, let's see if something regresses in the next image tests.

adamcstephens:
Is this where I'd check those tests? https://jenkins.linuxcontainers.org/job/lxd-test-images/restrict=master,type=vm/

stgraber:
Yeah, though they've been struggling a bit on other distros lately so we're unlikely to get a full green until @monstermunchkin comes back on Monday :)

PR #285, opened 2021-05-18
images/fedora: Fix fedora/34/cloud networking issues

monstermunchkin:


images/fedora.yaml
```diff
@@ -338,3 +338,17 @@ actions:
...
   types:
   - container
+
+- trigger: post-files
+  action: |-
+    #!/bin/sh
+    set -eux
+
+    # This fixes the networking issues
+    sed -ri 's#ProtectKernelTunables=.*#ProtectKernelTunables=yes#g' /etc/systemd/system/service.d/lxc.conf
```
stgraber, line 348: Why is this needed and how isn't that breaking networkd like it normally does?

monstermunchkin: No idea why this is needed, but it fixes all networking issues. It's not breaking networkd because cloud images use NetworkManager.

monstermunchkin: Actually, this causes some other issues. Marking it as a draft.

PR #287, opened 2021-05-19
images: Drop systemd changes

monstermunchkin:


stgraber:
Blocked on distrobuilder change, we need the systemd generator in place for this.

PR #291, opened 2021-05-26
VM: Add check that 9p config drive is exported readonly

tomponline:
Requires https://github.com/lxc/lxd/pull/8813


stgraber:
Similar to the comment in the LXD PR, this isn't something we can do, we have distros (and there will soon be more) which do not support 9p (centos and derivatives at least).

tomponline:
@stgraber have reworked this one for now to just check the 9p share is readonly.

PR #307, opened 2021-06-08
images: Fix systemd test

monstermunchkin:


stgraber:
For the sysctl ones, it'd be better if we had the generator override the sysctl unit to ignore failures when applying the sysctls, I believe systemd-sysctl has a command line flag for that.

monstermunchkin:
> For the sysctl ones, it'd be better if we had the generator override the sysctl unit to ignore failures when applying the sysctls, I believe systemd-sysctl has a command line flag for that.

The man page doesn't mention any such flag.

monstermunchkin:
@stgraber I removed the sysctl modifications in favor of https://github.com/lxc/distrobuilder/pull/453

images/opensuse.yaml
```diff
@@ -69,6 +69,13 @@ files:
 - path: /var/lib/dbus/machine-id
   generator: remove
 
+- path: /lib/modules-load.d/sg.conf
```
stgraber, line 72: I don't like us removing files that come from packages as those are likely to show up again in a package update.
Can we maybe figure out what processes those and instead disable or ignore failures on the systemd unit instead?

Containers in general cannot load kernel modules, so disabling the logic entirely should be fine.

monstermunchkin: Sure, that would be `systemd-modules-load.service`

PR #335, opened 2021-06-23
Network: Adds bridge firewall tests

tomponline:
Tests using snap with nftables and xtables drivers.

### DIFF
--- a/bin/test-lxd-network-bridge-firewall
+++ b/bin/test-lxd-network-bridge-firewall
@@ -1,0 +1,176 @@
+#!/bin/sh
+set -eu
+
+cleanup() {
+    echo ""
+    if [ "${FAIL}" = "1" ]; then
+        echo "Test failed"
+        exit 1
+    fi
+
+    echo "Test passed"
+    exit 0
+}
+
+FAIL=1
+trap cleanup EXIT HUP INT TERM
+
+# Wait for snapd seeding
+sleep 1m
+
+# Configure to use the proxy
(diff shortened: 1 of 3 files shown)


stgraber:
@tomponline The test is using a hardcoded device path (/dev/nvmeXYZ) but is run in virtual machines which will not have that.

tomponline:
> @tomponline The test is using a hardcoded device path (/dev/nvmeXYZ) but is run in virtual machines which will not have that.

Oops. Yes I changed that at the last minute from lantea to vm to avoid using lantea unnecessarily. Will change to loopback file.

tomponline:
@stgraber that should fix it now, I've added unmanaged bridge filtering checks too.

bin/test-lxd-network-bridge-firewall
```diff
@@ -0,0 +1,176 @@
...
+curl -s http://canonical-lxd.stgraber.org/config/snapd.sh | sh
+
+# Install LXD
+while :; do
+    [ ! -e /usr/bin/lxd ] && break
+    apt remove --purge lxd lxd-client --yes && break
+done
+snap install lxd --edge
+snap install jq
+#snap refresh lxd --channel=latest/edge
```
stgraber, line 31: This shouldn't be commented

tomponline: Doh yep, that was a workaround for the rate limiting from the snap store earlier.

tomponline: fixed

PR #353, opened 2021-07-14
images: Remove busybox-gzip only on tumbleweed for ppc64

monstermunchkin:


monstermunchkin:
Let's see if the build itself has been fixed, otherwise we need to do something different.

PR #356, opened 2021-07-19
Update images

monstermunchkin:


monstermunchkin:
@stgraber can we remove Ubuntu Xenial? It's not EOL (but neither was Trusty) but standard support has ended.

stgraber:
@monstermunchkin I'd keep it another 6 months at least as we still need it ourselves for testing, especially to test upgrading away from it

jenkins/jobs/image-apertis.yaml
```diff
@@ -19,8 +19,8 @@
         type: user-defined
         values:
         - 'v2019.7'
-        - 'v2020.4'
-        - 'v2021.0'
+        - 'v2020.5'
+        - 'v2021.1'
```
stgraber, line 23: Can we detect this somehow and have it be handled like the RHEL derivatives?
So instead having images for `v2020` and `v2021` and have those pick whatever the latest point release is?

Right now every time we bump, it means users need to change their scripts to pull the new version which isn't ideal.

monstermunchkin: Sure, I'll have distrobuilder find the latest release.

monstermunchkin: Actually, distrobuilder already supports this.

PR #367, opened 2021-07-22
bin/test-lxd-network: Stop then delete instances

tomponline:
Part of trying to identify the cause of intermittent deletion errors

### DIFF
--- a/bin/test-lxd-network
+++ b/bin/test-lxd-network
@@ -186,10 +186,15 @@ lxc list
 networkTests
 
 # Remove instances, leaving one VM and container for bridge hotplug test
-lxc delete -f v1-macvlan
-lxc delete -f c1-macvlan
-lxc delete -f v1-sriov
-lxc delete -f c1-sriov
+echo "=> Removing some instances"
+lxc stop -f v1-macvlan
+lxc delete v1-macvlan
+lxc stop -f c1-macvlan
+lxc delete c1-macvlan
+lxc stop -f v1-sriov
+lxc delete v1-sriov
+lxc stop -f c1-sriov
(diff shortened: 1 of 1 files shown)


bin/test-lxd-network
```diff
@@ -186,10 +186,15 @@ lxc list
...
-lxc delete -f c1-sriov
+echo "=> Removing some instances"
+lxc stop -f v1-macvlan
+lxc delete v1-macvlan
+lxc stop -f c1-macvlan
+lxc delete c1-macvlan
+lxc stop -f v1-sriov
+lxc delete v1-sriov
+lxc stop -f c1-sriov
+lxc delete c1-sriov
```
stgraber, line 197: If it's the same thing I've seen last time, then it won't help.

Last I was on the system and hit the issue, the problem was that `v1-sriov` or some other instance had file descriptors to files from `c1-macvlan` which then prevented the deletion of `c1-macvlan` until `v1-sriov` was deleted.

tomponline: Interesting, when/how did you ascertain that? I've tried doing that test in a loop repeatedly and wasn't able to recreate.

stgraber: I guess I got lucky, I just deployed lantea, transfered maas-kernel, ran `maas-kernel iommu`, then transfered `lxd-test-network` and ran that, hit the issue on first go.

tomponline: Did you see the file descriptors hanging around or just the mount error?

I'll try deploying lantea and see if I can reproduce there. What is the maas-kernel iommu thing you mentioned there?

PR #394, opened 2021-09-29
Add block storage test

monstermunchkin:


bin/test-lxd-block-storage
```diff
@@ -0,0 +1,107 @@
...
+lxc network create lxdbr0
+lxc profile device add default eth0 nic network=lxdbr0
+
+poolName="vmpool$$"
+
+for poolDriver in $poolDriverList
+do
+        echo "==> Create storage pool using driver ${poolDriver}"
+        if [ "${poolDriver}" = "dir" ]; then
+                lxc storage create "${poolName}" "${poolDriver}" volume.size=5GB
```
tomponline, line 48: Any particular reason for 5GB in this section, seems like for non-optimized block volumes this will just cause the tests to run slower compared to, say, 50MB? Is there a practical minimum size?

monstermunchkin: Yes, I went with 5GB because I got this error when creating the instance `Failed instance creation: Failed creating instance from image: Source image size (4294967296) exceeds specified volume size (50000000)`

monstermunchkin: done

tomponline (approved):
thanks!

PR #402, opened 2021-10-08
Update image-amazonlinux.yaml

ghost:
Original method of passing curl to sed was not supported by sh (I only tested in bash). 

With the original attempt, sh gives:
```
$ REL=$(sed -En 's/location:.*os-images\/([^\/]+)\/.*/\1/p' <(curl -sI https://cdn.amazonlinux.com/os-images/latest/)) && echo $REL
/bin/sh: 5: Syntax error: "(" unexpected
```

With the updated method, sh gives:
```
$ REL=$(curl -sI https://cdn.amazonlinux.com/os-images/latest/ | sed -En 's/location:.*os-images\/([^\/]+)\/.*/\1/p') && echo $REL
2.0.20210813.1
```

### DIFF
--- a/jenkins/jobs/image-amazonlinux.yaml
+++ b/jenkins/jobs/image-amazonlinux.yaml
@@ -36,7 +36,7 @@
         [ "${ARCH}" = "arm64" ] && ARCH="aarch64"
         
         REL=${release}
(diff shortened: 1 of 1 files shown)


theCiscoGeek:
I missed the echo that I had on there for testing. Will fix and resubmit.

PR #418, opened 2021-11-24
VM: Updates VM storage tests to use a waitVMAgent function to reduce sleep times

tomponline:
Also renamed and updated storage volume VM tests to accommodate restricted project testing.


simondeziel (approved):

tomponline:
@stgraber I believe the jenkins jobs will need to be refreshed after this is merged.

PR #419, opened 2021-11-24
bin/test-lxd-storage-volumes-vm: use simpler unit with df

simondeziel:
And only output device size. Use more precise matching with grep
while at it.

Signed-off-by: Simon Deziel <simon.deziel@canonical.com>

### DIFF
--- a/bin/test-lxd-storage-vm
+++ b/bin/test-lxd-storage-vm
@@ -87,7 +87,7 @@ do
         lxc exec v1 -- umount /srv
 
         echo "==> Checking VM root disk size is 10GB"
-        lxc exec v1 -- df -B1000000000 | grep sda2 | grep 10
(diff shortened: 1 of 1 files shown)


bin/test-lxd-storage-vm
```diff
@@ -87,7 +87,7 @@ do
         lxc exec v1 -- umount /srv
 
         echo "==> Checking VM root disk size is 10GB"
-        lxc exec v1 -- df -B1000000000 | grep sda2 | grep 10
+        lxc exec v1 -- df -B 1GB --output=size | grep sda2 | grep -w 10
```
tomponline, line 90: If you only output the size, won't that prevent the filter on `sda2`?

tomponline (approved):
@simondeziel LGTM, did the test run through OK?

PR #424, opened 2021-11-25
bin/test-lxd-*: use a subshell for waitSnapdSeed to isolate set +x

simondeziel:
Otherwise the undo the waitSnapdSeed `set -x` was affecting the global context.

Signed-off-by: Simon Deziel <simon.deziel@canonical.com>

### DIFF
--- a/bin/test-lxd-cgroup
+++ b/bin/test-lxd-cgroup
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -eu
 
-waitSnapdSeed() {
+waitSnapdSeed() (
   set +x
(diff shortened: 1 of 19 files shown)


stgraber:
You could run it in a sub-shell which would then let you do `set +x` without affecting the rest of the script.

stgraber:
Also you have a change that's not related to `set +x` in there.

simondeziel:
I took inspiration from Tomp's waitVMAgent function but didn't copy the important bit "()" vs "{}", thanks for catching this!

PR #428, opened 2021-11-30
Small simplifications

simondeziel:


bin/maas-run
```diff
@@ -123,8 +123,8 @@ done
 
 # Wait for SSH
 echo "==> Waiting for SSH to respond" >&2
+IP=$(maas maas machine read "${system_id}" | jq .ip_addresses[] -r | grep ":" | head -2)
 while :; do
-    IP=$(maas maas machine read "${system_id}" | jq .ip_addresses[] -r | grep ":" | head -2)
```
stgraber, line 127: That's intentional, we may be getting garbage at times so we want to fetch it again every time in case it changed to something better.

simondeziel: I see and the IP is garbage, `ssh ubuntu@IP && break` will avoid exiting due to `set -e`.

PR #446, opened 2022-01-24
bin/test-lxd-migration: Create snapshots after connectivity test.

### LINKED ISSUE #445: Ensure snapshots work after LXD migration
The `storage_volume` and `storage_volume_snapshots` tables may be error prone after a migration due to the `sqlite_sequence` table and some triggers (see https://github.com/lxc/lxd/pull/9807). 

We need to ensure that snapshots work correctly after a migration in case this comes up again.

markylaing:
Checks that snapshots work correctly after a migration. 

Closes #445

### DIFF
--- a/bin/test-lxd-migration
+++ b/bin/test-lxd-migration
@@ -147,3 +147,8 @@ for url in $(curl -s --unix-socket /var/snap/lxd/common/lxd/unix.socket "lxd/1.0
         FAIL=1
     fi
 done
+
+# Create some snapshots (catch possible error in sqlite_sequence)
+lxc snapshot c1
+lxc snapshot c1
+lxc snapshot c1


bin/test-lxd-migration
```diff
@@ -146,4 +146,9 @@ for url in $(curl -s --unix-socket /var/snap/lxd/common/lxd/unix.socket "lxd/1.0
         echo "FAIL-IPv6: ${name}"
         FAIL=1
     fi
+
+    # Create some snapshots (catch possible error in sqlite_sequence)
+    lxc snapshot "${name}"
+    lxc snapshot "${name}"
+    lxc snapshot "${name}"
```
stgraber, line 153: We don't need to try it on all containers, so I'd do it after the loop and just test on `c1` or something.

PR #457, opened 2022-02-22
test-lxd-network: Ensures that lxd-agent renames interfaces on VM startup.

### LINKED ISSUE #455: Update LXD network tests to include lxd-agent rename interfaces feature
The `agent.rename_interfaces` config key renames a VMs network interfaces on startup to match the names of attached NIC devices (see https://github.com/lxc/lxd/pull/9893).

We need to update the `bin/test-lxd-network` to include this feature.

markylaing:
Checks that the VMs default network interface is renamed to the device name on startup. 

This PR should not be merged until https://github.com/lxc/lxd/pull/9893 is merged.

Closes #455.


markylaing:
@tomponline @stgraber I've just set this to draft while https://github.com/lxc/lxd/pull/9893 is still being reviewed.

bin/test-lxd-network
```diff
@@ -221,6 +221,20 @@ sleep 10
 lxc list
 networkTests
 
+# Check bridged NIC type works with "agent.rename_interfaces".
+echo "=> Performing bridged NIC agent.rename_interfaces tests"
+lxc stop v1-physical
+lxc config set v1-physical agent.rename_interfaces=true
+lxc start v1-physical
+
+# Wait for lxd-agent to rename the interface.
```
stgraber, line 230: You're missing an actual sleep statement here.

markylaing: Fixed - not sure how that happened.

bin/test-lxd-network
```diff
@@ -221,6 +221,21 @@ sleep 10
...
+sleep 10
```
tomponline, line 231: This feels too short and at best will likely only work intermittently.
Can you use waitLXDAgent from https://github.com/lxc/lxc-ci/blob/master/bin/test-lxd-storage-disks-vm#L55 (copy it here)

bin/test-lxd-network
```diff
@@ -221,6 +221,21 @@ sleep 10
 lxc list
 networkTests
 
+# Check bridged NIC type works with "agent.rename_interfaces".
+echo "=> Performing bridged NIC agent.rename_interfaces tests"
+lxc stop v1-physical
```
tomponline, line 226: Please use stop -f for tests

bin/test-lxd-network
markylaing: @tomponline I've done this, but in doing so I've noticed there are a lot of these small functions duplicated across the scripts. Is there a reason we're doing this instead of just sourcing/importing them?

tomponline: I thought about this originally, and I'm not dead set against it. But I've found that when running these scripts locally to test they work before submitting the PR some of them require to be run on special hardware (thinking SR-IOV and GPU ones for example), so being able to just copy a single file to a specific machine (and quickly iterating) to run it was easier. So laziness basically :)

markylaing: Hmm... Yeah I guess it's not really a problem unless we find an issue with one of the functions and have to change them all and I guess we can make a decision about it at that point.

tomponline (approved):
Looks good. Does this work OK locally? cheers

markylaing:
> Looks good. Does this work OK locally? cheers

I tested my addition with a script locally
```
#!/bin/bash

lxc launch images:ubuntu/20.04 v1 --vm -c agent.rename_interfaces=true
while ! lxc exec v1 -- echo ""
do
  sleep 1
done

lxc exec v1 -- ip link show eth0

! lxc exec v1 -- ip link show enp5s0 || false
```
but I haven't ran this test in full locally because I wasn't certain how to handle the snap installation.

Looking at it again I think I can just comment out the snap stuff and run it against my local build.

tomponline:
Yeah its somewhat tricky, personally I have a separate little PC that is basically a fresh install of ubuntu 20.04 with the normal server kernel (not the HWE desktop kernel) and then I copy the script to it and run it, to let it install the snap and run end-to-end.
This is useful for testing SR-IOV networking and faster VM testing.

For the non-VM ones (not in this case) I spin up a VM (`lxc launch images:ubuntu/focal v1 --vm`) and run the script inside it.
You can even run VM ones inside a VM too, using VM nesting. You need to have the HWE kernel on both the host and the guest though (https://wiki.ubuntu.com/Kernel/LTSEnablementStack).

tomponline:
Otherwise it takes up to 24 hours to get the results of a change, which is no good when developing a new test where I may need to run it every few minutes :)

PR #477, opened 2022-03-23
Update ovn tests with network state and ACL logs

### LINKED ISSUE #449: Update OVN tests with latest features
We should update https://github.com/lxc/lxc-ci/blob/master/bin/test-lxd-network-ovn with tests for features added in:

- https://github.com/lxc/lxd/pull/9813
- https://github.com/lxc/lxd/pull/9850

markylaing:
Ensures that `lxc info` container the MTU and MAC address for an OVN network. 
Ensures that the `lxc network acl show-log` command does not fail.

Closes #449


bin/test-lxd-network-ovn
```diff
@@ -1014,6 +1014,7 @@ ovn_peering_tests() {
     lxc exec ovn1 -T -n --project=ovn1 -- ping -c1 -6 -W5 "${ovn2NICIPv6}"
     lxc exec ovn2 -T -n --project=ovn2 -- ping -c1 -4 -W5 "${ovn1NICIPv4}"
     lxc exec ovn2 -T -n --project=ovn2 -- ping -c1 -6 -W5 "${ovn1NICIPv6}"
+    lxc network acl show-log ovn1 --project=ovn1
```
markylaing, line 1017: @tomponline I initially was going to grep for something in the output of this but there doesn't seem to be any output? Is that right?

tomponline: There should only be output if one of the rules that have logging enabled is matched.

simondeziel (approved):

bin/test-lxd-network-ovn
```diff
@@ -72,6 +72,10 @@ ovn_basic_tests() {
     sleep 2
     lxc network list | grep ovn-virtual-network
 
+    echo "==> Check we can get the network state (MAC address and MTU are reported in network info)"
+    lxc network info ovn-virtual-network | grep -P 'MAC address: [0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}'
+    lxc network info ovn-virtual-network | grep -P 'MTU: \d+'
```
tomponline, line 77: Can you add a check for the active chassis member please?

bin/test-lxd-network-ovn
stgraber: Right, you'll need to generate a bit of traffic to make that work.

markylaing: Ah I missed the `state=logged` configuration.

markylaing: I've set the state of one of the ACL rules to logged and checked that all the fields are the expected values with jq.

tomponline (approved):

PR #487, opened 2022-04-06
images/debian: systemd-networkd fixes

monstermunchkin:

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1104,10 +1104,25 @@ actions:
     #!/bin/sh
     set -eux
 
+    # Disable networkd service; this prevents the container from getting 2 IPv4 addresses on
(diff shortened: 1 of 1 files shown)


images/debian.yaml
```diff
@@ -1107,7 +1122,7 @@ actions:
     # Enable networkd socket
     systemctl enable systemd-networkd.socket
   types:
-  - containers
```
tomponline, line 1110: @monstermunchkin is "containers" valid at any time? If not should the template validation not have failed the image build entirely with this old value?

monstermunchkin: It's invalid, and we don't have template validation yet.

tomponline: Oh :( I created https://github.com/lxc/distrobuilder/issues/614 for that :)

PR #491, opened 2022-04-08
bin/test-lxd-images: Skip CGroupV1 images

monstermunchkin:
Some images require CGroupV1 which cannot be satisfied by the
testserver. Therefore, these distributions cannot be tested and need
to be excluded.

### DIFF
--- a/bin/test-lxd-images
+++ b/bin/test-lxd-images
@@ -82,6 +82,10 @@ lxc profile device add default root disk pool=default path=/
 # Create the instances
 echo "=> Creating the instances"
 for i in $(lxc image list images: | grep x86_64 | grep "${TYPE}" | grep "more" | awk '{print $2}'); do
+    # Skip images requiring CGroupV1 as the testserver uses CGroupV2
+    if [ "${TYPE}" = "CONTAINER" ] && lxc image info "images:${i}" | grep -q "requirements.cgroup: v1"; then
+        continue
+    fi
     name=$(echo "${i}" | sed -e "s/\//-/g" -e "s/\.//g")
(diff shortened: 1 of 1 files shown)


bin/test-lxd-images
```diff
@@ -82,6 +82,10 @@ lxc profile device add default root disk pool=default path=/
 # Create the instances
 echo "=> Creating the instances"
 for i in $(lxc image list images: | grep x86_64 | grep "${TYPE}" | grep "more" | awk '{print $2}'); do
+    # Skip Amazon Linux as it requires CGroupV1 which the testserver cannot satisfy
+    if [ "${i}" = "amazonlinux" ]; then
+        continue
+    fi
```
stgraber, line 88: This will also fail for centos/7, oracle/7, spingdalelinux/7 and ubuntu/16.04.

Instead of the hardcoded skip, it'd be better to look at:
```
stgraber@dakara:~$ lxc image show images:centos/7 | grep requirements.cgroup
  requirements.cgroup: v1
```

bin/test-lxd-images
```diff
@@ -82,6 +82,10 @@ lxc profile device add default root disk pool=default path=/
 # Create the instances
 echo "=> Creating the instances"
 for i in $(lxc image list images: | grep x86_64 | grep "${TYPE}" | grep "more" | awk '{print $2}'); do
+    # Skip images requiring CGroupV1 as the testserver uses CGroupV2
+    if lxc image info "images:${i}" | grep -q "requirements.cgroup: v1"; then
+        continue
+    fi
```
stgraber, line 88: This should only be done for containers

monstermunchkin: done

PR #499, opened 2022-04-21
Improve reliability of test-lxd-images

simondeziel:


bin/test-lxd-images
```diff
@@ -189,6 +189,20 @@ for url in $(lxc query "/1.0/instances" | jq -r .[]); do
...
+            echo "FAIL: systemd clean: ${name}"
+            # Show the systemd failures
+            lxc exec "${name}" -- systemctl --failed
+            FAIL=1
+        else
+            echo "PASS: systemd clean: ${name}"
+        fi
+    else
+        echo "SKIP: systemd clean: ${name}"
+    fi
```
stgraber, line 215: I don't mind the re-order but this won't really change anything in the end as we don't skip the other tests on failure.

bin/test-lxd-images
```diff
@@ -189,6 +189,14 @@ for url in $(lxc query "/1.0/instances" | jq -r .[]); do
         continue
     fi
 
+    # Wait for instance to be ready to execute commands
+    for i in $(seq 30); do
+        if lxc exec "${name}" -- true 2>/dev/null; then
+            break
+        fi
+        sleep 2
+    done
```
stgraber, line 198: How is that needed when we've already gone through a `sleep 1m` after the last instance has been started?

We also wait 10s after every 3 instance starts, so for large sets of instances (like our usual 100+), this means we'll have waited well over 6min by the time we run the tests here.

bin/test-lxd-images
```diff
@@ -199,6 +199,8 @@ for url in $(lxc query "/1.0/instances" | jq -r .[]); do
 
     # Systemd cleanliness
     if lxc exec "${name}" -- sh -c "type systemctl" >/dev/null 2>&1; then
+        # Allow more time to boot if needed
+        lxc exec "${name}" -- systemctl --quiet is-system-running || sleep 30
```
stgraber, line 203: Similar as before, we'll normally hit this 5min after initial startup.
If the instance still isn't up, something is very wrong and it deserves a failure here :)

bin/test-lxd-images
```diff
@@ -221,6 +221,8 @@ for url in $(lxc query "/1.0/instances" | jq -r .[]); do
 
     if [ -z "${address}" ]; then
         echo "FAIL: No network interface: ${name}"
+        # Show the network state
+        lxc query "${url}/state" | jq -r ".network"
```
stgraber, line 227: Any reason not to show a simple `lxc info` here? that tends to be slightly more readable :)

PR #504, opened 2022-04-25
Fix Debian bullseye/cloud networking

### LINKED ISSUE #503: Image debian/bullseye/cloud gets 2 IPv4 addresses on startup
Hi,

When launching a Debian bullseye with the cloud variant:

```shell
lxc launch images:debian/bullseye/cloud mycontainer
```

... both `cloud-init` and `systemd-networkd.service` run a DHCP request and eth0 gets 2 IPv4 addresses.

It seems this issue has already been addressed in #487 but the fix is not working. Indeed, since `systemd-networkd.socket` is enabled, it seems `systemd-network.service` is run as a kind of dependency.

I guess the fix is to not enable `systemd-networkd.socket` for the bullseye cloud variant. Am I correct?
In the coming days, I will try to run distrobuilder to test my assumption.

Cheers,
Albin

monstermunchkin:
This resolves #503.

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1123,6 +1123,7 @@ actions:
 
     # Disable networkd service; this prevents the container from getting 2 IPv4 addresses on
     # startup. This only happens on debian/bullseye/cloud images.
+    systemctl disable systemd-networkd.socket
     systemctl disable systemd-networkd.service
   types:
   - container
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1142,7 +1142,6 @@ actions:
   types:
   - container
   releases:
-  - bullseye
   - bookworm
   - sid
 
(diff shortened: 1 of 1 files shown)


images/debian.yaml
```diff
@@ -1141,7 +1142,6 @@ actions:
   types:
   - container
   releases:
-  - bullseye
```
albinou, line 1144: Not tested, but I am afraid that debian/bullseye (variant default) will not get an IPv4 because `systemd-networkd.socket` is not enabled.

albinou (commented):
Thanks for this fix!
I have a comment though.

images/debian.yaml
monstermunchkin: You're right.

monstermunchkin: Fix: https://github.com/lxc/lxc-ci/pull/505

Issue #490, opened 2022-04-08
Update OVN tests to run on Ubuntu Jammy


tomponline:
The OVN daily tests currently depend on a feature that was removed from OVN :(

See https://github.com/ovn-org/ovn/issues/124

We should update our OVN tests to not rely on this feature by adding a static route manually.

tomponline:
Working on this. Even with specific fix i had in mind several other steps are failing intrrmittently.

tomponline:
Working on this now.

tomponline:
Looks like OVN in Ubuntu Jammy is broken - I keep getting ovn-controller using 100% CPU intermittently and the network not working properly.

tomponline:
Reopened as although we have added the required changes to get the tests to run on Jammy, there is an issue in the Jammy OVN version that is causing ovn-controller to use high CPU and stop functioning when networks are being deleted, causing the tests to fail intermittently.

PR #521, opened 2022-05-11
test: Add Jammy support for OVN network tests

tomponline:
Requires https://github.com/lxc/lxd/pull/10393

Closes https://github.com/lxc/lxc-ci/issues/490

### DIFF
--- a/bin/test-lxd-network-ovn
+++ b/bin/test-lxd-network-ovn
@@ -676,6 +676,7 @@ ovn_forward_tests() {
     # Add forward to u1 and test working.
     lxc network forward create ovn-virtual-network 192.0.2.1 target_address="${U1_IPV4}"
     lxc network forward create ovn-virtual-network 2001:db8:1:2::1 target_address="${U1_IPV6}"
+    sleep 2
 
     # Check forward is working by pinging it.
     # Relies on static route (above) rather than neighbour adverts see https://github.com/ovn-org/ovn/issues/124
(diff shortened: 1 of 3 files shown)


tomponline:
@stgraber this can be merged, and we'll see how reliable (or not) this is on the daily tests.
It is possible to get it to complete all the way through, but the OVN version in Jammy is buggy and somtimes the ovn-controller process uses 100% CPU and the network doesn't operate. I'm struggling to identify what causes it as the logs don't contain anything different when it fails compared to a successful run. However it does appear to be related to deleting an OVN network.

tomponline:
@stgraber suffice to say I strongly recommended sticking with Focal for your OVN hosts for now.

stgraber:
But I'm running the jammy version of OVS and OVN and a kernel more recent than jammy so the rest of userspace being focal shouldn't matter much.

tomponline:
> But I'm running the jammy version of OVS and OVN and a kernel more recent than jammy so the rest of userspace being focal shouldn't matter much.

Ah OK if you're running a custom version then will be interesting to see if you encounter the issues in production, or whether its just the behavior of the tests that is triggering it intermittently.

tomponline:
@stgraber do you have a ppa I can validate with?

tomponline:
What I do see in the logs that is different than the Focal OVN version is:

```
2022-05-11T19:21:41.270Z|00070|if_status|WARN|Dropped 2 log messages in last 50 seconds (most recently, 45 seconds ago) due to excessive rate
2022-05-11T19:21:41.270Z|00071|if_status|WARN|Trying to release unknown interface cr-lxd-net9-lr-lrp-ext
```

stgraber:
ppa:stgraber/stgraber.net is what I run on my production servers

PR #531, opened 2022-05-24
jenkins/jobs: Drop i386 images for pld

simondeziel:


jenkins/jobs/image-pld.yaml
```diff
@@ -1,5 +1,3 @@
-- job:
-    name: "image-pld"
```
stgraber, line 2: What happened here?

simondeziel: Whoa, probably bogus keystrokes in vim, fixed now.

jenkins/jobs/image-pld.yaml
```diff
@@ -1,6 +1,5 @@
 - job:
-    name: "image-pld"
-    concurrent: false
+    name: "image-pld"    concurrent: false
```
stgraber, line 2: Still not quite right, you're missing a line break here :)

simondeziel: Now the diff looks good.

PR #548, opened 2022-06-14
Fix AlmaLinux 9

liberodark:
Fix Boot issue with AlmaLinux 9 LXC on PVE 7.2

Related to : https://forum.proxmox.com/threads/lxc-almalinux-9-issue.110348/#post-477778

Tested and work on Fresh install.

### DIFF
--- a/images/almalinux.yaml
+++ b/images/almalinux.yaml
@@ -454,6 +454,7 @@ actions:
     printf "[main]\ndhcp=dhclient" > /etc/NetworkManager/conf.d/dhcp-client.conf
 
     systemctl enable NetworkManager.service
+    systemctl mask systemd-firstboot.service
   types:
   - container
   variants:


stgraber:
What's happening with that systemd unit exactly?

In general, we prefer having systemd workarounds applied through the systemd generator in distrobuilder so that should the rootfs be used outside of a LXC container, that unit doesn't remain masked/disabled.

@monstermunchkin

michael-manley:
@stgraber From what I can tell the service just stays in the activating state, causing the rest of the system not to start. I found that when I stop the service, RHEL 9 Derivatives boot fine. Same issue happens on CentOS Stream 9 as well.

liberodark:
Hi, @stgraber 

service stay active : 

```
systemctl status systemd-firstboot
● systemd-firstboot.service - First Boot Wizard
     Loaded: loaded (/usr/lib/systemd/system/systemd-firstboot.service; static)
    Drop-In: /run/systemd/system/service.d
             └─zzz-lxc-service.conf
     Active: activating (start) since Wed 2022-06-15 07:28:51 UTC; 36s ago
       Docs: man:systemd-firstboot(1)
   Main PID: 71 (systemd-firstbo)
      Tasks: 1 (limit: 256041)
     Memory: 1.2M
     CGroup: /system.slice/systemd-firstboot.service
             └─71 systemd-firstboot --prompt-locale --prompt-timezone --prompt-root-password

Notice: journal has been rotated since unit was started, output may be incomplete.
```

That cause some issue console not work correcly only black screen & you can login with console.
network is also not work if systemd-firstboot is active.

```
 ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0@if19: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether d6:f8:f4:4a:ce:c3 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

stgraber:
@monstermunchkin can we sort this out in distrobuilder?

stgraber:
Dealt with in distrobuilder

PR #549, opened 2022-06-14
Fix CentOS 9

liberodark:
Fix Boot issue with CentOS 9 LXC on PVE 7.2

Related to : https://forum.proxmox.com/threads/lxc-almalinux-9-issue.110348/#post-477778

Tested and work on Fresh install.


stgraber:
What's happening with that systemd unit exactly?

In general, we prefer having systemd workarounds applied through the systemd generator in distrobuilder so that should the rootfs be used outside of a LXC container, that unit doesn't remain masked/disabled.

@monstermunchkin

stgraber:
Dealt with in distrobuilder

PR #564, opened 2022-06-29
images/centos: Use ipv6_dhcpv6-stateless for 9-Stream/cloud

monstermunchkin:
This fixes an issue where the instance would not get an IPv6 which also
caused DNS not to work.


monstermunchkin:
> Hrm, this doesn't seem right.

How so?

> As for IPv6, the default operation mode for our images is SLAAC, so we would expect the container images to be configured to use SLAAC configuration by default.

`ipv6_dhcpv6-stateless` is SLAAC. See the [docs](https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v1.html#subnet-ip).

tomponline:
> `ipv6_dhcpv6-stateless` is SLAAC. See the [docs](https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v1.html#subnet-ip).

Well not quite, it is a superset that uses SLAAC for address and route configuration, and DHCPv6 for DNS configuration.

In that context `ipv6_slaac` is pure RA SLAAC. IPv6 RA can also include DNS info btw.

But looking into the change set more closely it seems that the commit message isn't telling the whole story, because it looks like you're also configuring DHCPv4 when `network_mode` isn't `link-local`.  So does that cover the DHCPv4 scenario?

So I have a few of questions:

1. What is `network_mode` in this context? Where does it come from?
2. Did you evaluate using `ipv6_slaac` and find it was not suitable compared to `ipv6_dhcpv6-stateless` somehow? Does DHCPv4 not configure the DNS servers when using `ipv6_slaac` mode?
3. How does this functionality compare to Ubuntu/Debian images in terms of them using pure-SLAAC vs stateless DHCPv6. It would be good to get an understanding as to what our images are doing in different scenarios.

Thanks

monstermunchkin:
> 1. What is `network_mode` in this context? Where does it come from?

This can actually be removed, see [here](https://discuss.linuxcontainers.org/t/lxd-first-class-cloud-init-support/12559/9?u=monstermunchkin).

> 2. Did you evaluate using `ipv6_slaac` and find it was not suitable compared to `ipv6_dhcpv6-stateless` somehow?

I just did, and it works just as well.

> Does DHCPv4 not configure the DNS servers when using `ipv6_slaac` mode?

It does:
```
NetworkManager[129]: <info>  [1656512282.3196] dhcp4 (eth0):   address 10.75.225.5
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   plen 24 (255.255.255.0)
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   gateway 10.75.225.1
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   lease time 3600
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   hostname 'centos-9-stream-cloud'
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   nameserver '10.75.225.1'
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0):   domain name 'lxd'
NetworkManager[129]: <info>  [1656512282.3197] dhcp4 (eth0): state changed new lease, address=10.75.225.5
```

> 3. How does this functionality compare to Ubuntu/Debian images in terms of them using pure-SLAAC vs stateless DHCPv6. It would be good to get an understanding as to what our images are doing in different scenarios with a view to making them behave consistently as possible.

Ubuntu images don't specify an IPv6 subnet, only `dhcp`. Let's see if CentOS 9-Stream can work that way as well.

monstermunchkin:
It turns out, we can just drop the whole `content` thing. I don't recall why we needed to set `content` in the past but I'll check whether older CentOS releases can drop that as well.

tomponline:
Excellent thank you.

PR #567, opened 2022-07-01
jenkins/jobs/almalinux: Add v9/s390x

monstermunchkin:


stgraber:
We already tried that and had to revert it.

stgraber:
AlmaLinux's s390x requires a more recent IBM z mainframe than what we're using to build our images.

monstermunchkin:
Oh right, I totally forgot about that.

PR #616, opened 2022-09-14
images/ubuntu: Add systemd-resolved to kinetic

monstermunchkin:


tomponline (commented):
Can we make it default installed for Ubuntu variants and then disable it for selected older ubuntu versions, rather than have to enable it for every new version?

monstermunchkin:
> Can we make it default installed for Ubuntu variants and then disable it for selected older ubuntu versions, rather than have to enable it for every new version?

Nope. `systemd-resolved` is part of the `systemd` package for older versions and therefore is always installed. `kinetic` and newer releases have a separate `systemd-resolved` package.

PR #632, opened 2022-10-20
bin/test-lxd-cluster: Fix clustering test

tomponline:
By installing lxd snap package in each cluster member. Rather than installing in first and then copying it.


tomponline:
Shouldnt need this if we fix the error in snap package:

```
Oct 20 21:26:37 c1 lxd.daemon[1558]: /snap/lxd/23832/commands/daemon.start: 489: CURRENT_PID: parameter not set
```

PR #642, opened 2022-11-16
images: Add trailing / to avoid HTTP permanent redirects

simondeziel:


simondeziel:
@monstermunchkin I noticed a bunch of mirror/url supports HTTPS but do not redirect HTTP->HTTPS. Would you be OK with us using HTTPS everywhere it's supported?

tomponline:
Seems sensible. Although I'm guessing we also check the signatures anyway.

simondeziel:
My trust in GPG is rather low these days but yeah, we are checking it AFAIK but belt and braces ;)

tomponline (approved):

PR #653, opened 2022-11-24
Shellcheck

simondeziel:


simondeziel:
@stgraber, I believe the GitHub Action will require you to do something in the repo config.

FYI, I picked this action over others because it is referenced on the upstream shellcheck repo https://www.shellcheck.net/wiki/GitHub-Actions

stgraber:
@simondeziel I don't think we'll have to do anything, it just needs to be merged in the main repo before it kicks in.

PR #654, opened 2022-11-25
devlxd: Fix test

monstermunchkin:


bin/test-lxd-devlxd-vm
```diff
@@ -91,7 +91,8 @@ lxc exec v1 -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0/devi
 lxc exec v1 -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0/config | jq
 lxc exec v1 -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0/meta-data | grep -q '#cloud-config'
 
-lxc restart -f v1
+# Use --timeout instead of --force otherwise the filesystem will get corrupted.
+lxc restart v1 --timeout=60
```
tomponline, line 95: Can you add this:

```
lxc restart v1 --timeout=60 || lxc restart v1 -f
```

So it always restarts, but gives it a good chance to cleanly restart.

bin/test-lxd-devlxd-vm
```diff
@@ -107,7 +108,10 @@ echo "==> Checking devlxd is not working"
 
 ! lxc exec v1 -- curl -s --unix-socket /dev/lxd/sock http://custom.socket/1.0 || false
 
-lxc restart -f v1
+# Let some time pass before restarting the VM again, otherwise it won't know that it needs to restart.
+sleep 5
```
tomponline, line 112: Why is the sleep 5 needed here before the restart, the VM should be up and running by this point because you used `waitVMAgent` above.

monstermunchkin: That should be the case, but unfortunately it isn't. Adding a short sleep, let's it restart.

tomponline: Can you try execing a `sync` command inside the VM before doing a forceful restart to see if that fixes this, without needing the graceful restart and sleeping.

monstermunchkin: Adding `sync` fixes all our problems. I'll update the PR.

tomponline (approved):

PR #658, opened 2022-11-30
image/openeuler: Update mirror

monstermunchkin:


pkking:
maybe duplicates #661

PR #660, opened 2022-11-30
LXD Cloud end to end tests with Juju and MAAS

simondeziel:


bin/test-lxd-cloud
```diff
@@ -0,0 +1,85 @@
...
+# Juju and MAAS creds
+eval "$(curl -s http://canonical-lxd.stgraber.org/config/juju.sh)"
+
+# Build LXD Cloud
+cd /root
+[ -d /root/.ssh ] || mkdir -m 0700 /root/.ssh
+if ! [ -f /root/.ssh/id_ed25519 ]; then
+    touch /root/.ssh/id_ed25519
+    chmod 0600 /root/.ssh/id_ed25519
+    echo "$SSH_PRIV_KEY" > /root/.ssh/id_ed25519
```
stgraber, line 69: Where does that come from? juju.sh?

I think it'd be better to move to a separate file so we don't accidentally expose lxd-cloud clone credentials to jobs that don't need it.

PR #675, opened 2022-12-13
jenkins/jobs: maas-run shouldn't need sudo

simondeziel:
Not using sudo should make it easier to cancel jobs.


tomponline:
Why was it used originally?

simondeziel:
The origin wasn't exactly clear to @stgraber but he confirmed that `jenkins` has all it needs from sourcing `/lxc-ci/etc/config` so root *shouldn't* be needed.

tomponline:
Excellent!

PR #681, opened 2023-01-03
jenkins/jobs: Add Linux Mint 21.1 ("Vera")

monstermunchkin:


jenkins/jobs/image-mint.yaml
```diff
@@ -40,7 +41,7 @@
         suite=bionic
         if [ "${release}" = "ulyana" ] || [ "${release}" = "ulyssa" ] || [ "${release}" = "uma" ] || [ "${release}" = "una" ]; then
             suite=focal
-        elif [ "${release}" = "vanessa" ]; then
+        elif [ "${release}" = "vanessa" ] || [ "${release}" = "vera" ]; then
```
tomponline, line 44: Can this be an else?

monstermunchkin: No. The default is `bionic` for `tara`, `tessa`, and `tina`. If this were an `else`, those `t`-releases would also use `jammy`.

tomponline: Sounds like its worth re-organising so that the older exceptions are explicitly stated, and the newer ones just default to jammy.

monstermunchkin: I don't think it's worth re-organizing now. The `t`-releases will go away in April 2023, then we can switch to an `else`.

PR #684, opened 2023-01-04
test-lxd-gpu-mig: Drop quotes around $UUIDS

monstermunchkin:
This removes the quotes around $UUIDS as they lead to UUIDx each being
a list of UUIDs instead of a single one.


tomponline:
@simondeziel is there a way to achieve this result whilst appeasing shellcheck?

PR #700, opened 2023-01-23
dqlite-benchmark: maas-run requires artifacts.tar.gz

MathieuBordere:
Signed-off-by: Mathieu Borderé <mathieu.bordere@canonical.com>

### DIFF
--- a/bin/test-dqlite-benchmark
+++ b/bin/test-dqlite-benchmark
@@ -73,5 +73,6 @@ sudo apt-get install -y dqlite-tools
 run_benchmark
 sleep 2
 run_benchmark --disk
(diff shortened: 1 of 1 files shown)


bin/test-dqlite-benchmark
```diff
@@ -73,5 +73,6 @@ sudo apt-get install -y dqlite-tools
 run_benchmark
 sleep 2
 run_benchmark --disk
+tar -zcf /home/ubuntu/artifacts.tar.gz dqlite-benchmark--disk.tar.gz dqlite-benchmark.tar.gz
```
monstermunchkin, line 76: Is there a typo and should read `dqlite-benchmark-disk.tar.gz` instead (double hyphen)?

MathieuBordere: The double hyphen is correct, I pass the argument to the script directly to the filename, it's maybe not super pretty.

PR #702, opened 2023-01-23
build_distro: Enable debug logging

monstermunchkin:
Looking at the image build logs, what we have been seeing a lot lately
is:

```
Error: Instance not found
Error: Failed checking instance exists "local:distrobuilder-<uuid>": Instance not found
```

This PR is an attempt to find the underlying issue.

### DIFF
--- a/bin/build-distro
+++ b/bin/build-distro
@@ -18,11 +18,22 @@ shift
 TARGET=${1}
 shift
 
+success=0
+
 cleanup() {
     lxc delete --force "${CNAME}"
+
+    # Only show debug log if build is unsuccessful
+    if [ "${success}" -eq 0 ]; then
+        snap logs -n=1000 lxd
+    fi
 }
 
 trap cleanup EXIT HUP INT TERM
+
+# Set up debug logging
+snap set lxd daemon.debug=true
+systemctl reload snap.lxd.daemon
 
 # Create the container
 lxc copy "cache-distrobuilder-${ARCH}" "${CNAME}"
@@ -85,3 +96,5 @@ lxc exec "${CNAME}" -- tar -cf - -C /root/build/ . | tar -xvf - -C "${TARGET}"
 
 [ -n "${SUDO_UID:-}" ] && chown "${SUDO_UID}" -R "${TARGET}"
 [ -n "${SUDO_GID:-}" ] && chgrp "${SUDO_GID}" -R "${TARGET}"
+
+success=1


tomponline:
That seems a bit odd, can we instead investigate why the instance that has just been copied (the `lxc copy` command has returned without error) is not found when trying to start it. That seems like a possible bug in LXD and should be fixed there rather than working around it in the client (in this case distrobuilder).

tomponline (changes requested):
Lets fix the issue in LXD rather than workaround it.

monstermunchkin:
I don't believe this is an issue in LXD. We see this mostly on ARM builds but I'm unable to produce the error. An alternative solution would be to add `sync` between copying and starting the container.

tomponline:
The Instance not found error is coming from the DB. So I can understand better, what is the setup of LXD in this scenario? And why do you think a `sync` call would help change what is in the DB?

monstermunchkin:
> So I can understand better, what is the setup of LXD in this scenario?

It's not a special setup, just the normal snap. The build container is created in `bin/build-image-distrobuilder`.

> And why do you think a `sync` call would help change what is in the DB?

`sync` occasionally solves mysterious problems, and since I have no idea what is causing this issue I'll try anything.

tomponline:
OK, try setting up debug logging to and see what is the order of things happening.
It certainly doesn't seem right for this to be happening in the snap.

bin/build-distro
```diff
@@ -18,12 +18,23 @@ shift
...
 shift
 
+success=0
+
 cleanup() {
     lxc delete --force "${CNAME}"
+
+    # Only show debug log if build is unsuccessful
+    if [ "${success}" -eq 0 ]; then
+        snap logs -n=1000 lxd
```
tomponline, line 28: I did not know about that command, thanks for the tip!

bin/build-distro
```diff
@@ -18,12 +18,23 @@ shift
...
+    fi
 }
 
 trap cleanup EXIT HUP INT TERM
 
+# Set up debug logging
+snap set lxd daemon.debug=true
+snap restart lxd
```
tomponline, line 36: `sudo systemctl reload snap.lxd.daemon` would be less impactful, would that help?

monstermunchkin: But `reload` won't restart the LXD daemon with debugging enabled. If at all we should use `restart` or not?

tomponline: sudo systemctl reload snap.lxd.daemon does restart the LXD daemon, but without restarting the instance too.
I always use this way after I enable debugging.

monstermunchkin: OK, I've updated it.

tomponline (approved):

stgraber:
No no no, don't do that.

`build-distro` runs directly on the builddXX, not in VMs or anything, you absolutely DO NOT want to mess with global LXD settings on there.

Pushing a revert.

stgraber:
And going to fix them all manually now.

tomponline:
Ah OK thought this was a distrobuilder specific test script.

tomponline:
@monstermunchkin @stgraber confirmed that it was the `lxc copy` command that was failing and not the `lxc start` command, and the 2nd error was the TRAP at the top of the script. So the `lxc start` command was never even called.

PR #707, opened 2023-02-14
test: Attach custom non-block volume to VM

monstermunchkin:


tomponline (approved):
Thanks. Would we expect lxd-agent to mount those in the VM? should we test for that too?

monstermunchkin:
> Thanks. Would we expect lxd-agent to mount those in the VM? should we test for that too?

What exactly do you mean? Are you saying we should test whether the mountpoint is actually being created?

tomponline:
Yeah I'm trying to remember if lxd-agent mounts those file shares inside the guest automatically? If so we could test for that to ensure they are actually working.

PR #722, opened 2023-04-04
Drop Go 1.17

simondeziel:
Version 1.17 is EOL and doesn't match the version shipped in
Ubuntu LTS releases (Focal ships 1.13 and Jammy 1.18).


tomponline:
Your sign off name is wrong. Also why are we dropping this? The commit doesn't give a reason.

PR #744, opened 2023-05-18
switch to mirrors.ocf.berkeley.edu

pkking:
As openEuler's first North American mirror site setup, i'd like give it a shoot :)

### DIFF
--- a/images/openeuler.yaml
+++ b/images/openeuler.yaml
@@ -4,7 +4,7 @@ image:
     openEuler {{ image.release }}
 source:
   downloader: openeuler-http
-  url: https://mirrors.aliyun.com/openeuler/
+  url: https://mirrors.ocf.berkeley.edu/openeuler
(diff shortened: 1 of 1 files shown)


images/openeuler.yaml
```diff
@@ -4,7 +4,7 @@ image:
     openEuler {{ image.release }}
 source:
   downloader: openeuler-http
-  url: https://mirrors.aliyun.com/openeuler/
+  url: http://mirrors.ocf.berkeley.edu/openeuler
```
simondeziel, line 7: That Berkeley mirror seems to support HTTPS too.

simondeziel (changes requested):

images/openeuler.yaml
pkking: Done

simondeziel (approved):

PR #748, opened 2023-06-02
images/ubuntu: Fix permission for netplan config

monstermunchkin:


tomponline (approved):

images/ubuntu.yaml
```diff
@@ -807,5 +807,11 @@ actions:
   types:
   - vm
 
+- trigger: post-files
+  action: |-
+    #!/bin/sh
+    set -eux
+    [ -f /etc/netplan/10-lxc.yaml ] && chmod 0600 /etc/netplan/10-lxc.yaml
```
tomponline, line 814: Why don't we set the correct perms when its creates rather than later though?

lukasz0617: Hi, I'm having the same issue with netplan and used your code to set permissions. I'm still getting "Cannot call openvswitch: ovsdb-server.service is not running" . Am I doing something wrong here? are there any further steps needed?

monstermunchkin: @lukasz0617 can you confirm that `10-lxc.yaml` has the correct permission set in the container?

lukasz0617:
Hi @monstermunchkin ,

I did set it up as 0600. The file name in my case is different but I'm sure permissions are set.

image

I only get this issue when i run 'sudo netplan apply' and it is only the message that is the issue. The IP configuration o the ens160 adapter are there and I can communicate out of the server.  I haven't touched linux for a while so my skills are a bit rusty but by the looks of it there is no service called ovsdb-server.service.

Is this service needed? Do I need to install it to get rid of the message?

Thanks

images/ubuntu.yaml
loptap: First time creating new instance for a few weeks and I am NOW also seeing

> "Permissions for /etc/netplan/90-lxc.yaml are too open. Netplan configuration should NOT be accessible by others. WARNING:root:Cannot call Open vSwitch: ovsdb-server.service is not running"

with image 22.04 (unchanged apart from any automatic updating). Note that 90-lxc.yaml was a pushed file and the permissions were indeed too open - I simply `chmod 0600 /etc/netplan/90-lxc.yaml` - and then `netplan try` accepted the config without issue but still reported 

> "Cannot call openvswitch: ovsdb-server.service is not running"

The LXD set up has been in place for a while. Doesn't seem to create any issues, but clearly not right?

PR #754, opened 2023-06-13
images/debian: Fix networkd in bookworm containers

kdrag0n:
Default Debian bookworm container images boot with the eth0 interface down, so networking does not work. Starting `systemd-networkd` manually fixes the network.

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1198,6 +1198,7 @@ actions:
   - default
   releases:
   - bullseye
+  - bookworm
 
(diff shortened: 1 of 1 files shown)


monstermunchkin:
@kdrag0n Does this also work with bookworm/cloud container images?

kdrag0n:
Upon further investigation, the activation socket is listening on netlink and LXD triggers it because it starts the container with eth0 already up. Fixed on my side by doing the same thing.

PR #757, opened 2023-06-19
images/archlinux: Delete [community] repo from pacman.conf

monstermunchkin:


images/archlinux.yaml
```diff
@@ -652,6 +652,9 @@ actions:
     #!/bin/sh
     set -eux
 
+    The [community] repo has moved into [extra]
+    sed -ri '/\[community\]/,+2d' /etc/pacman.conf
```
simondeziel, line 656: It looks like the comment is missing a `#`. As for the `sed`, I'd maybe use anchors to avoid matching on random potential occurrences:

```
sed -i '/^\[community\]$/,+2d' /etc/pacman.conf
```

simondeziel (changes requested):

simondeziel (approved):
Thanks

PR #761, opened 2023-06-26
test-lxd-storage-volumes-vm: Test custom ISO volumes

monstermunchkin:


simondeziel (approved):
I have not reviewed the PR adding support for ISO volumes but was wondering if those are meant to be mounted concurrently on multiple machine and also readonly?

If yes to the above, would it be possible to test this part as well?

monstermunchkin:
> If yes to the above, would it be possible to test this part as well?

Yes, we could test that. And thanks for the review!

tomponline:
@monstermunchkin can we add iso volume tests to the main test suite, including attaching to a container (like a block volume), then this daily tests can just be whether it is actually usable as a cdrive on a vm.

monstermunchkin:
> can we add iso volume tests to the main test suite, including attaching to a container (like a block volume), then this daily tests can just be whether it is actually usable as a cdrive on a vm?

I've done that already. Though custom ISO volumes cannot be attached to containers (which is tested).

tomponline:
OK cool, why can they not be attached btw?

monstermunchkin:
> OK cool, why can they not be attached btw?

It's a VM feature for now, just like custom block volumes.

tomponline:
> 

Yeah, although I thought one could attach a block volume to a container, is that still not possible?

monstermunchkin:
> > 
> 
> Yeah, although I thought one could attach a block volume to a container, is that still not possible?

Nope, still not possible.

```
$ lxc storage volume attach default vol1 c1
Error: Failed add validation for device "vol1": Custom block volumes cannot be used on containers
```

bin/test-lxd-storage-volumes-vm
```diff
@@ -121,25 +123,54 @@ do
...
 	# Both volumes can be attached at the same time. The VM will have /dev/sdb and /dev/sdc.
 	# It doesn't matter which volume is which device as they contain the same content.
 	lxc storage volume attach "${poolName}" vol2 v1
 	lxc storage volume attach "${poolName}" vol3 v1
 
+	echo "==> Attach custom ISO volumes to VM"
+	lxc storage volume attach "${poolName}" vol5 v1
+	lxc storage volume attach "${poolName}" vol6 v1
+	lxc storage volume attach "${poolName}" vol5 v2
+	lxc storage volume attach "${poolName}" vol6 v2
```
simondeziel, line 148: That tests multi-attach which is good. Since the ISO volume allows concurrent readonly mounts, would it also make sense to test that out by `mount`'ing the same volume(s) in both `v1` and `v2` and then maybe `umount` in one and make sure it still works in the other. Presumably this `mount`/`umount` dance will be more relevant when containers will be supported but wouldn't hurt to test it on VMs too.

monstermunchkin: I changed it to:
```shell
lxc storage volume detach "${poolName}" vol5 v1
lxc exec v2 -- /bin/sh -c 'mount /dev/sr0 /mnt && [ $(cat /mnt/foo) = foo ] && ! touch /mnt/foo && umount /mnt'
```

simondeziel: If I followed correctly, the `touch` makes sure the FS is mounted readonly so that's good.

Sorry, I was not clear with what I meant by "concurrent mounts"; let me try with code:

```
# concurrent readonly ISO mounts
# shellcheck disable=2016
lxc exec v1 -- /bin/sh -c 'mount /dev/sr0 /mnt && [ $(cat /mnt/foo) = foo ] && ! touch /mnt/foo'
# shellcheck disable=2016
lxc exec v2 -- /bin/sh -c 'mount /dev/sr0 /mnt && [ $(cat /mnt/foo) = foo ] && ! touch /mnt/foo'
lxc exec v1 -- umount /mnt
lxc exec v2 -- umount /mnt
```

tomponline:
@monstermunchkin please can you apply this change to https://github.com/canonical/lxd-ci/blob/main/tests/test-lxd-storage-volumes-vm as well? Thanks

PR #763, opened 2023-06-26
Small tweaks to `bin/build-image-distrobuilder`

simondeziel:
@monstermunchkin I've dropped `subversion` assuming no distro use it anymore, please correct me if that isn't the case. Thanks.


bin/build-image-distrobuilder
```diff
@@ -75,10 +75,32 @@ apt-get update --yes
 
 # Install dependencies
 apt-get install --yes --no-install-recommends \
-    build-essential btrfs-progs git ca-certificates gcc libc6-dev debootstrap \
-    xz-utils squashfs-tools gpg dirmngr gpg-agent rsync kpartx python3 \
-    subversion patch build-essential gawk libncurses-dev unzip qemu-utils \
-    gdisk dosfstools snapd zstd
+    btrfs-progs \
+    build-essential \
+    build-essential \
```
monstermunchkin, line 84: Duplicate

bin/build-image-distrobuilder
```diff
@@ -11,9 +11,10 @@ trap cleanup EXIT HUP INT TERM
 
 ARCH="$1"
 TARGET="$2"
+DISTRO="jammy"
```
monstermunchkin, line 14: That makes sense :)

monstermunchkin (changes requested):

monstermunchkin:
> I've dropped subversion assuming no distro use it anymore, please correct me if that isn't the case.

TBH I don't know if this is the case. However, should tests fail because of this, we'll just re-add it.

tomponline (approved):

monstermunchkin (approved):

PR #765, opened 2023-06-27
images(ubuntu): moved lxd-growpart service to non-desktop variant vm

MrDaGree:
Theres been a few times I prefer to use linuxcontainer ubuntu images vs ubuntu remote images but I've noticed when using the non-desktop variant vm types the disk root partition does not grow to the correct size set in my lxd profile.

My solution was to move the `lxd-growpart.service` out of the section that only applies for the destkop variant


simondeziel (approved):
Good catch! I didn't notice `cloud-guest-utils` was installed on all VMs. Thanks!

simondeziel:
@MrDaGree would you mind adding a `Signed-off-by` tag to your commit? Otherwise our DCO check will fail once it runs.
`git commit -s --amend` should do the right thing.

tomponline (commented):
Do we need to exclude this from the `/cloud` variants though? As they have cloud-init which provides this functionality. Otherwise will they conflict?

monstermunchkin:
We'll need the same change for openSUSE and Arch Linux.

monstermunchkin:
> Do we need to exclude this from the `/cloud` variants though? As they have cloud-init which provides this functionality. Otherwise will they conflict?

Just to be sure we should exclude cloud images, i.e. use

```
variants:
- default
```

MrDaGree:
Sounds good. I’ll get the additional requests added later today!

monstermunchkin:
@MrDaGree Could you please make a separate commit for each distro?

MrDaGree:
Sorry about that, updated for each distro to be its own commit

PR #766, opened 2023-06-27
Remove support for 2.0 (EOL)

simondeziel:
@stgraber is `bin/test-lxd-migration` still in use? I couldn't find where if it is.


stgraber:
The Jenkins job for it has been dropped recently as there shouldn't be any normally supported .deb LXD system left with 18.04 going EOL.

PR #767, opened 2023-06-28
Fixes for build-image-distrobuilder

monstermunchkin:


simondeziel (approved):
Thanks for fixing this. It's a bit of a shame that I did that same error with DISTRO twice already :/

PR #773, opened 2023-07-13
Fix openwrt dnsmasq/dhcp on boot

mgziminsky:
Use noop instead of comment to remove calls to `procd_add_jail` so usages inside a shell pipeline don't result in syntax errors.

https://github.com/openwrt/openwrt/commit/936df715de3d33947ce38ca232b05c2bd3ef58f1 added a call to `procd_add_jail_mount` as part of a short-circuit test. Using a comment breaks the script.

Tested in a proxmox VM

### DIFF
--- a/images/openwrt.yaml
+++ b/images/openwrt.yaml
@@ -349,7 +349,7 @@ actions:
     #!/bin/sh
 
     # Disable process isolation to make dnsmasq work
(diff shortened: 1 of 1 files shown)


imEgo:
lxc.cap.drop: sys_admin makes dnsmasq not working, this cap dropped in openwrt.common.conf, after commented this, dnsmasq working on PVE 8.1.4 with privileged Container

PR #778, opened 2023-08-12
Add support for slackware

slapcat:
- supports verisons 15.0 and current


slapcat:
Distrobuilder [PR#766](https://github.com/lxc/distrobuilder/pull/766) has been merged and is available in `latest/edge`, so this PR/job should be valid now.

PR #779, opened 2023-08-15
Add cowsql benchmark job

freeekanayaka:
This job/script for now only runs the `raft-benchmark disk` tool to benchmark raw disk sequential write performance.

More benchmarks will be added in the future.

### DIFF
--- a/bin/test-cowsql-benchmark
+++ b/bin/test-cowsql-benchmark
@@ -133,7 +133,8 @@ while :; do
 done
 
 # Setup dependencies
-sudo apt-get install -y "${PACKAGES}"
+# shellcheck disable=SC2086
+sudo apt-get install -y ${PACKAGES}
 
 curl -s $BENCHER_RELEASES_URL | \
     grep "browser_download_url.*_$(dpkg --print-architecture).deb" | \
--- a/bin/test-cowsql-benchmark
+++ b/bin/test-cowsql-benchmark
@@ -1,0 +1,154 @@
+#!/bin/sh
+set -eu
+
+export DEBIAN_FRONTEND=noninteractive
+DIR=$(pwd)
+DATA_DIR=$DIR/cowsql-benchmark
+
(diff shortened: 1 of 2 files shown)


freeekanayaka:
[Here](https://bencher.dev/console/projects/raft/perf?benchmarks_per_page=8&testbeds_per_page=8&testbeds_page=1&benchmarks_page=1&branches_per_page=8&branches_page=1&reports_per_page=4&reports_page=1&clear=true&metric_kind=latency&branches=33ef1617-322b-454a-bf3f-b3d04e3c7558&testbeds=536270a2-9532-47f9-8941-5e4a1fb0edde%2C1e55a996-556b-483c-9a98-f5354a026e00&tab=testbeds&benchmarks=efa3430f-b408-477b-9193-a90c8c77ecf8%2C5333ffa7-46b6-4d70-a253-c30f58bc26d2%2C9464855b-c85f-44a2-8853-ccb03cf57276&key=true&range=version) you can see some data comparing the latency of io_uring, pwritev2 and io_submit (KAIO), running with direct I/O and fdatasync-ing each write:

The very erratic part of the graph on the left shows data generated running the benchmark with GitHub runners, while the more stable part on the right is data from lxc's lantea. Notice how much more predictable it is! That will help to spot small improvements or regressions.

Interestingly io_submit latency seems slightly lower than io_uring on this specific hardware, perhaps both because the Ubuntu 22.04 kernel is already a bit behind (io_uring progress is fast) and because there are some powerful io_uring knobs to use that the benchmark currently doesn't take advantage of.

I've added a lot of combinations of file systems, buffer sizes and write engines, but I'll probably dial that matrix down a bit moving forward. This is mostly to see if data match expectations.

stgraber:
The test scripts is already run through `sudo` by `maas-run`, so you should be able to drop a bunch of `sudo` calls in there :)

PR #785, opened 2023-09-01
Run cowsql job using the latest Zabbly kernel

freeekanayaka:
This enables us to track performance improvements or regressions in latest kernel versions (in particular in `io_uring`-related code).

### DIFF
--- a/bin/maas-kernel
+++ b/bin/maas-kernel
@@ -160,6 +160,24 @@ case "$i" in
         apt-get dist-upgrade --yes
       ;;
 
+      zabbly)
+        echo "===> Installing the current zabbly kernel"
+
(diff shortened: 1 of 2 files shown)


freeekanayaka:
Not sure if it's fine to place the APT pub key blob inline in the `maas-kernel` script, happy to change that if there's a better place to put it.

stgraber:
@freeekanayaka you could do `curl https://pkgs.zabbly.com/key.asc -o /etc/apt/keyrings/zabbly.asc` to keep it a bit shorter

freeekanayaka:
> @freeekanayaka you could do `curl https://pkgs.zabbly.com/key.asc -o /etc/apt/keyrings/zabbly.asc` to keep it a bit shorter

Much nicer indeed, thanks. That's done.

Issue #786, opened 2023-09-05
NixOS images


adamcstephens:
Hello. I've recently improved the NixOS support for building LXC container images and LXD virtual machine images. I originally investigated using distrobuilder to do so, but it was mostly trying to fit a square peg into a round hole. Nix already provides all the necessary build infrastructure that works with NixOS, and distrobuilder makes way too many assumptions about building a typical Linux distro that do not apply to Nix. 

My question here is, would it be possible to consume the images that the Nix CI builds so that they can be served through the linuxcontainers image server? Effectively the builder would only need to retrieve the latest tarballs/images that are already being produced.

It looks like NixOS supported was requested way back in #38 and the guidance was to use distrobuilder, but I'm hopeful we can find a compromise that does not meet that guidance. Thanks!

stgraber:
It's not impossible but it's definitely not our preference.

What's the issue with distrobuilder exactly?
There already are quite a few images which are pretty thin repack performed by distrobuilder where a mostly complete rootfs is downloaded from an existing source, gets validated and then only some config files get put in place or in more extreme cases, just the image metadata files.

adamcstephens:
NixOS expects and enforces that configuration is produced by the defined system configuration. For example all of `/etc/systemd` is symlinks to read-only files in the Nix store, `/nix/store/....`. So even the act of generating configuration should be done through NixOS tooling and not externally. None of the distrobuilder assumptions around package installation, network configuration, systemd overrides, etc, work with how NixOS expects the system to be configured. Distrobuilder also makes many assumptions about the existence of /usr, /sbin, /lib, that do not exist on a NixOS system.

I asked for some guidance a year ago [on discourse](https://discuss.linuxcontainers.org/t/nixos-image-building/14999) but ended up not getting any feedback. One thing I remember (and see in my posts from the time) is that the container images we produce through the NixOS CI aren't even chrootable [until this script snippet executes](https://github.com/NixOS/nixpkgs/blob/63ff76729b7359cd01e4ce09455b6af6c45f7286/nixos/modules/virtualisation/lxc-container.nix#L28-L36) on first boot.

These are just some of the challenges and are due to NixOS being fundamentally incompatible with assumptions that are safe on almost any other distro.

adamcstephens:
[...]

PR #802, opened 2023-12-27
nixos: add VM images from hydra

adamcstephens:
I think this is what @stgraber was suggesting in https://github.com/lxc/lxc-ci/issues/786 , but if not please let me know.

Guarded to unstable for now. We will backport the necessary hydra change if everything looks good here, and then this can be updated to support both.


adamcstephens:
Not sure why this test failed. https://jenkins.linuxcontainers.org/job/test-image/architecture=amd64/4053/console

adamcstephens:
Actually I think I know the root cause. It’s a path issue when running exec commands on a VM.

adamcstephens:
This should fix the problem but will take a few days to be in our images after merge. https://github.com/NixOS/nixpkgs/pull/277235

PR #808, opened 2024-01-03
Add Network compatibility on debian

liberodark:
This changes.

Fix possibility to work on pct with PVE.
`ifupdown2` help to set/reload ip configuration also for what `ifupdown2` and not `ifupdown` because `ifupdown` is now deprecated on next debian release.
This change `/etc/network` fix possibility to add compatibility for add or not `interfaces` file in this folder.

Build OK / Test OK

Best Regards

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1183,6 +1183,7 @@ packages:
   - packages:
     - dbus
     - systemd-resolved
+    - ifupdown2
     action: install
     early: true
     releases:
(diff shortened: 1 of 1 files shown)


stgraber:
Hmm, ifupdown2 pulls in python3 which then adds more than 10% to the image size.

I don't see a good reason to include this when systemd-networkd works fine for our needs.

liberodark:
On my side that no work but is possible to set only ifupdown is tiny software.
This is long issue for somme users so i think is a good workaround.

stgraber:
28MB when the current image unpacks to less than 230MB, isn't "small" :)

PR #809, opened 2024-01-03
Fix Duplicate IP debian

liberodark:
This fix duplicate IP on containers when you setup ip fixe.

### DIFF
--- a/images/debian.yaml
+++ b/images/debian.yaml
@@ -1339,8 +1339,10 @@ actions:
     #!/bin/sh
     set -eux
 
-    # Enable networkd socket
-    systemctl enable systemd-networkd.socket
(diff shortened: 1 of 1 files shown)


stgraber:
There is no such thing as duplicate IPv4 addresses with our Debian images.

I suspect the reason why you're seeing duplicate IPv4 addresses is because you're somehow using both ifupdown2 (per your other pull request) as well as our default systemd-networkd.

Since the addition of iupdown2 isn't really an option, this change also feels wrong.

liberodark:
Is only when you have ip fixe

stgraber:
If you are using a pre-defined IPv4 address, then you're expected to turn off the DHCP client in the instance.

PR #811, opened 2024-02-03
images/gentoo: installkernel >=20 updates

juippis:
- add "grub" use flag with openrc (I forgot this last time),
 - "systemd-kernel-install" was settled with "systemd-boot" instead on systemd.

NOTE: I couldn't boot the systemd variant, but openrc fully works. I'm gonna need some time to study systemd-boot as I've never dealt with it before. However the `systemd-boot` use flag looks final, so this should still be merged **if** systemd-boot is to be used in future with systemd vm variants.

### DIFF
--- a/images/gentoo.yaml
+++ b/images/gentoo.yaml
@@ -572,7 +572,10 @@ actions:
     systemd-machine-id-setup
     bootctl install --no-variables --esp-path=/boot/efi/
 
-    echo "sys-kernel/installkernel dracut systemd systemd-kernel-install" >> /etc/portage/package.use/installkernel
+    echo "root=/dev/sda2" >> /etc/kernel/cmdline
+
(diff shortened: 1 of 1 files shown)


juippis:
Just throwing my observation about systemd-boot here - not sure if it's relevant or not: The /boot files are installed on the "EFI" partition (/dev/sda1) while with openrc they are installed on rootfs (/dev/sda2). No separate /boot is mounted on either case.

juippis:
Force-pushed an update. Adding 
`echo "root=/dev/sda2" >> /etc/kernel/cmdline`
makes the systemd vm work for me locally - not sure if `/dev/sda2` is something that'd only work here, but I think that's a hint of sort what causes the issue. Please take it from here :)

PR #822, opened 2024-04-02
Fixed hardcoded root partition device name sda2

nanjj:
The root partition may not be /dev/sda2, for example if set the boot disk as the second disk, the root partition device name will be /dev/sdb2. It may be /dev/vda2 if not using san disk. The fix figured out the root partition device UUID and set the UUID as the root, the root partition device UUID keeps no change.

Depends on lxc/distrobuilder#830

### DIFF
--- a/images/almalinux.yaml
+++ b/images/almalinux.yaml
@@ -409,7 +409,7 @@ actions:
     target="$(readlink -f /etc/grub2-efi.cfg)"
     grub2-mkconfig -o "${target}"
 
-    sed -i "s#root=[^ ]*#root=/dev/sda2#g" "${target}"
+    sed -i "s#root=[^ ]*#root=${DISTROBUILDER_ROOT_UUID}#g" "${target}"
   types:
   - vm
   releases:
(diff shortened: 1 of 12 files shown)


stgraber:
Hmm, I like the idea behind the change but I'm not sure that the implementation is going to be safe. We frequently build multiple images in parallel on the same system which means that any kind of wildcard pattern around /dev/loop*p2 is going to be racy.

stgraber:
Yeah, I'd actually prefer we change distrobuilder to expose env variables, that'd be far more reliable.

/dev is usually a devtmpfs or can get mounted as one by a package during the build process, devices aren't namespaced so loop devices will show up in all containers/namespaces that also use devtmpfs.

nanjj:
The code I mentioned here: 
https://github.com/lxc/distrobuilder/blob/main/distrobuilder/vm.go#L164
So I do not think all loop devices in host will show up in all containers.

stgraber:
Note line 149, the logic you're pointing out is fallback logic for the images that do not have a full /dev already available. The code as it's written actually shows that distrobuilder was meant to support images building with a full /dev instance.

nanjj:
> Note line 149, the logic you're pointing out is fallback logic for the images that do not have a full /dev already available. The code as it's written actually shows that distrobuilder was meant to support images building with a full /dev instance.

OK, I changed my mind, now the fix is based on PR lxc/distrobuilder#830

stgraber:
Going to wait a day before merging that one so we can be sure all the builders get the new distrobuilder first.

PR #829, opened 2024-05-24
alpine: Add 3.20

dimitry-unified-streaming:
Alpine 3.20 was released 2024-05-22:
https://www.alpinelinux.org/posts/Alpine-3.20.0-released.html


dimitry-unified-streaming:
Btw, if this requires resources that aren't available, you might consider dropping Alpine 3.16 images, since that is EOLed since 2024-05-23.

stgraber:
Thanks. I've now removed 3.16 as we don't keep EOL releases around.

PR #834, opened 2024-06-24
Keep ubuntu network settings consistent

nanjj:
Related with PR lxc/distrobuilder#859.


stgraber:
We're not going to do that. It will literally break everyone's existing deployment logic and goes against the distro's default behavior.

PR #837, opened 2024-07-25
Gentoo template update

winterheart:
Remove GPG workaround as redundant.

Remove changing RSYNC_MIRROR commands as effectively noop. Add ability to use binary packages when possible. Remove redundant portage files after installation (source files, binary packages, portage tree).

### DIFF
--- a/images/gentoo.yaml
+++ b/images/gentoo.yaml
@@ -488,14 +488,7 @@ actions:
     #!/bin/sh
     set -eux
 
-    # Workaround GPG issues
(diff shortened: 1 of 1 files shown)


stgraber:
Why is the mirror configuration being removed? 

We can't use the primary mirror as we quickly hit some rate limits and get blocked from accessing it.

winterheart:
If you mean RSYNC_MIRROR, it's unused in image building, instead of this there webrsync is used with source.url setting.

PR #877, opened 2025-04-01
test-image: fix typo causing the script to fail silently

lnussbaum:
Context: https://discuss.linuxcontainers.org/t/sharing-directory-with-debian-13-vm-does-not-work/23361


monstermunchkin (approved):
Good catch!

PR #884, opened 2025-04-30
alpine: enable ppc64el and s390x

fossdd:
Both archtectures are supported for all non-EOL Alpine versions. The baseline for s390x is z196 and therefore works with the z13 builder.

### DIFF
--- a/jenkins/jobs/image-alpine.yaml
+++ b/jenkins/jobs/image-alpine.yaml
@@ -14,6 +14,8 @@
         - arm64
         - armhf
         - riscv64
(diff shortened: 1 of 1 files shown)


stgraber:
We don't have ppc64el or s390x hardware in our lab.

fossdd:
I see. IBM offers s390x virtual machines to open source projects, would you be interested in this? https://community.ibm.com/zsystems/form/l1cc-oss-vm-request/

They offer z15 hardware which would re-enable other distros like Alma aswell.

PR #905, opened 2025-07-20
rename incus ovn test projects

The127:
Renames the incus projects in the incus ovn tests to make them distinct from the network names.


stgraber:
@The127 looks like the commit is missing the usual signed-off-by line.

Otherwise I'm happy to merge this even before the Incus fix lands as I don't have a problem with extending our test suites to cover what should be working setups, if that fails, it then makes it easy to validate the fix :)

PR #909, opened 2025-08-19
Add image generation for AlmaLinux 10

lnussbaum:
The container image works (at least for `default`, I did not test `cloud`).

Regarding the VM image :
- it boots fine (but I don't understand why the approach used for v9 cannot be adapted for v10 -- I took the code from @MichaelStauber's version which works)
- however the agent does not start

At the end of boot, console shows:
```
[    9.036680] virtio-fs: tag <config> not found
[   14.317760] virtio-fs: tag <config> not found
[   19.396814] virtio-fs: tag <config> not found
[   24.647583] virtio-fs: tag <config> not found
[   29.897760] virtio-fs: tag <config> not found
[   35.140829] virtio-fs: tag <config> not found
[   40.395432] virtio-fs: tag <config> not found
[   45.644483] virtio-fs: tag <config> not found
[   50.893362] virtio-fs: tag <config> not found
```

The changes should not affect AlmaLinux 8 or 9.

### DIFF
--- a/images/almalinux.yaml
+++ b/images/almalinux.yaml
@@ -490,17 +490,22 @@ actions:
   action: |-
     #!/bin/sh
     set -eux
+
+    # Track down the kernel
     kver=$(basename /boot/initramfs-*.img | sed -r 's#initramfs-(.+)\.img#\1#' | head -n1)
     if [ -z "$kver" ]; then
       echo "No initramfs found in /boot!" >&2
       exit 1
     fi
+
     kernel_path=$(find /boot -name "vmlinuz-${kver}" || true)
     if [ -z "$kernel_path" ]; then
       echo "Kernel image /boot/vmlinuz-${kver} not found!" >&2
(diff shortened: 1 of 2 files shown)


stgraber:
@lnussbaum can you edit your commits to include the `Signed-off-by` line?

For the agent, it's expected that 9p won't work on RHEL derivatives, for those, you need to do:
```
incus config device add NAME agent disk source=agent:config
```

That then provides the same data through a cd-rom drive and should get the agent behaving.
Would be great if you could test that with your images.

MichaelStauber:
> We'll just want to wait for @MichaelStauber to confirm he's fine with his grub logic being included here.

Yes, the changes @lnussbaum did look good. And good catch about ...

`incus config device add NAME agent disk source=agent:config`

... as our Incus GUI adds that automatically to created instances, so that we can use Spice on EL9 and EL10 Incus nodes.

PR #941, opened 2025-12-08
nixos: write unstable serial file

adamcstephens:
Pulled the date command from bin/build-distro. This won't match the serial in the metadata tarball, but maybe that's ok?


adamcstephens:
I see, it needs to be written to $WORKSPACE.

PR #975, opened 2026-04-16
ubuntu: Add resolute

dimitry-unified-streaming:
Adds Ubuntu 26.04 (Resolute Raccoon).

I'm unsure if this is everything needed to properly add a Jenkins job for these images, so please point me in the right direction if I need to update other things.

### DIFF
--- a/bin/jenkins-import-images
+++ b/bin/jenkins-import-images
@@ -36,6 +36,7 @@ ALIASES = {
     "ubuntu/noble": ["ubuntu/24.04"],
     "ubuntu/plucky": ["ubuntu/25.04"],
     "ubuntu/questing": ["ubuntu/25.10"],
+    "ubuntu/resolute": ["ubuntu/26.04"],
 }
 
 REQUIRE_CGROUP1 = [
--- a/images/ubuntu.yaml
+++ b/images/ubuntu.yaml
@@ -432,6 +432,7 @@ packages:
     - jammy
     - noble
     - questing
+    - resolute
     action: install
 
   - packages:
@@ -469,12 +470,14 @@ packages:
     releases:
     - noble
     - questing
+    - resolute
 
   - packages:
     - systemd-timesyncd
     action: install
     releases:
     - questing
+    - resolute
 
   - packages:
     - grub-efi-amd64-signed
@@ -602,6 +605,7 @@ actions:
   releases:
   - noble
   - questing
+  - resolute
 
 - trigger: post-packages
   action: |-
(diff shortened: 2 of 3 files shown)


stgraber:
I don't think we need the combination filter part. That was a temporary hack due to questing having some issues. 

Rest looks fine but we won't be merging that until after 26.04 LTS is actually out, so next week at earliest.

DimitryAndric:
Great, thanks. I've been testing the betas and the docker images for some time, and while it had a period of rather severe instability (as in, apt upgrade and the system's bricked :) ), it now seems to be quite OK. I hope Canonical manages to release it on 2026-04-23!

DimitryAndric:
It appears that https://jenkins.linuxcontainers.org/job/image-ubuntu/4632/ shows failures for armhf and riscv64.

The armhf ones both error out with:

```text
+ ln -sf ../run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
ln: Already exists
Error: Failed to run post-packages: exit status 1
```
which is an interesting failure: I would have thought `-f` overwrites symlinks unconditionally? But there is probably something strange going on.

The riscv64 ones both error out with:

```text
W: Failure trying to run: chroot "/root/build/rootfs" /bin/true
W: See /root/build/rootfs/debootstrap/debootstrap.log for details
```

Since the Jenkins job doesn't show the contents of debootstrap.log, it's not really clear what goes wrong...

stgraber:
I believe I read somewhere that Ubuntu 26.04 has changed riscv64 compatibility quite significantly so that may just be broken for good and need to skip it.

stgraber:
Pushed some changes now, we'll see if that helps

DimitryAndric:
Pretty strange, now it gives:

```
+ rm -f /etc/resolv.conf
rm: cannot remove '/etc/resolv.conf': Device or resource busy
```

Not sure what is keeping the file occupied. Maybe systemd-resolved?

stgraber:
Ah, yeah, easy enough to fix, I'll deal with that.

DimitryAndric:
It's a little weird that this doesn't happen on e.g. amd64 and aarch64. What is so special about armhf, apart from it being the only 32 bit architecture?

DimitryAndric:
It seems to have worked now: https://jenkins.linuxcontainers.org/job/image-ubuntu/4638/ . What is still needed to make those images available on the main images.linuxcontainers.org site?

stgraber:
It's failing its tests because of some netplan failure.

PR #978, opened 2026-04-26
test-image: use build-name plugin to expose image name

adamcstephens:
I don't have a jenkins server to try this on, but my working theory is the [build-name](https://jenkins-job-builder.readthedocs.io/en/latest/wrappers.html#wrappers.build_name) wrapper could maybe clarify which test is for which image. Assuming this theory is correct, and this ends up in the rss feed title, I can filter in my reader down to just nixos.

This will obviously require the plugin if not already installed. Thoughts?

### DIFF
--- a/jenkins/jobs/test-image.yaml
+++ b/jenkins/jobs/test-image.yaml
@@ -49,3 +49,5 @@
     wrappers:
     - ansicolor:
         colormap: css
+    - build-name:
+        name: '#${BUILD_NUMBER} ${image}'


stgraber:
We can give that a try, let's see.

adamcstephens:
@stgraber was this applied? I assumed it would get picked up automatically, but nothing seems to have changed.

stgraber:
It should have, all the test runners pull from this git repo every 30min

PR #980, opened 2026-05-05
openwrt: build type=container,vm by default

maertsen:
With the release of Openwrt 25.12.3, and thanks to the work of @rietbergenm, it should now be possible to build both container and vm images by default going forward. 

Change the default, with a container-only exception for 24.10.


maertsen:
Thanks for merging @stgraber!

It seems like this patch missed something, because Jenkins appears not to be building VM images yet, and still runs the old version for 25.12 on a daily basis.

At what point should the new config start getting used?

PR #994, opened 2026-05-25
nixos: only repackage 25.11 images

herbetom:
The test to use the official images (#940) has been a success.

Therefore this can done for all images going forward.

At any point after 30.06.2026 we can clean this up while dropping 25.11.

cc @adamcstephens


adamcstephens:
Thanks!

